PR #2, opened 2017-09-04
Adjust label location, tabindex and accept button.

fileproxy:
Just started using this tool and love it. These are some changes to form controls.


amibu01:
Thanks!

PR #14, opened 2019-03-19
login form should accept return to submit the values

blckshdw:
The username/password textboxes should accept Return to submit the login request


goaaats:
Thanks!

PR #17, opened 2019-03-24
auto-set-default-to-latest-expansion

Jaehyuk-Lee:
I think this is better to set default expansion.


goaaats:
This already is the case, at least for me - can someone else confirm?

blckshdw:
I had a default value as well

Jaehyuk-Lee:
I mean, you don't have to change default value when another expansion releases.

sslesar:
Some people have starter edition or trial.

goaaats:
That should be fine since I clamp the maximum expansion level based on what square provides: https://github.com/goaaats/FFXIVQuickLauncher/blob/master/XIVLauncher/XIVGame.cs#L44

Really, I might as well get rid of that option, since all it does it change the title screen.

Jaehyuk-Lee:
Ah, I didn't know that does only changes title screen. I think it's ok to remove it. I thought it was existing because some change happens in launcher with other expansions.

PR #27, opened 2019-05-03
Clean up weird dimensions and fix indentation in MainWindow.xaml

Outfenneced:
Removes weird negative values in the grid of the login area (large negative values in the margin). Just figured I should clean it up a little since I was looking in the area. You may want to double-check my math, but before and after look identical to my eye.

New on top:
New vs old


goaaats:
Thanks! Not really used to WPF, guess this works :)

PR #30, opened 2019-05-26
Better OTP UI Behavior & Resized Username/PW Boxes

Mortalitas:
Instead of triggering on the 6th key, One Time Passwords now require a Return or Enter to be submitted.

Ok / Cancel buttons have been added to the One Time Password prompt.

Username & Password boxes resized by 1px for the sake of aesthetics, as text was being partially cut off on Windows 7.


Mortalitas:
Ignore the Delete MainWindow bits, Git is hard, mmk?

goaaats:
Looks good, thank you!

PR #34, opened 2019-05-30
Add files via upload

Genesis-Nova:
Changed the auto-login administrator check to check if the shift key is being held instead.
Creating pull request as asked.


goaaats:
Thanks!

PR #67, opened 2019-07-09
Nested settings control

samueltstephenson:
Moved settings window to a user control and nested it in main window, displays with a material style slide transition on click.


goaaats:
Did this in https://github.com/goaaats/FFXIVQuickLauncher/commit/f8afcf63cc1fd69c4ace9fb1c41fd0380f9adf91

PR #101, opened 2019-10-04
Only copy settings

pigsflew:
Skip, for instance, the item sorts, gearsets, and gold saucer stuff.


pigsflew:
This isn't tested yet, my local machine is missing some linked assemblies...

Issue #118, opened 2019-11-17
No fitting error message for boot version conflict


blckshdw:
System.AggregateException: One or more errors occurred. ---> System.Exception: Cannot verify Game version, Boot is outdated.
   at XIVLauncher.Game.XivGame.RegisterSession(OauthLoginResult loginResult)
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at XIVLauncher.Game.XivGame.Login(String userName, String password, String otp, Boolean isSteamServiceAccount, Boolean useCache)
   at XIVLauncher.Windows.MainWindow.<StartGame>d__14.MoveNext()
---> (Inner Exception #0) System.Exception: Cannot verify Game version, Boot is outdated.
   at XIVLauncher.Game.XivGame.RegisterSession(OauthLoginResult loginResult)
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()<---

4.1.0.0
89ca295-dirty
Context: Login
Microsoft Windows NT 6.2.9200.0
True
True
False
True
False
English

[]

Zyian:
Sounds like you need to run the original launcher first to get all of the updates. If the boot executable has an update it needs to get updated first so that quick launch can get to it

goaaats:
Exactly, I'll add a better error message for this

goaaats:
Boot patch error is now handled in the same way as game patch as of https://github.com/goaaats/FFXIVQuickLauncher/commit/d736a6c2c33a17e7853ec064746b8aa45bb5e9c2

PR #122, opened 2019-12-08
Changed margins and heights of listview

### LINKED ISSUE #111: Settings -> Addon: List of Addons Cut's Off/Does Not Scroll
See Screenshot: 

image

Zyian:
This should resolve #111 

There's also a change to improve the error about updating boot possibly for #118 ?

Current behavior
image

Initial View
image

Scrolled down
image

### DIFF
--- a/XIVLauncher/Windows/SettingsWindow.xaml.cs
+++ b/XIVLauncher/Windows/SettingsWindow.xaml.cs
@@ -459,5 +459,10 @@ namespace XIVLauncher.Windows
                 Log.Error(exc, "Couldn't check dalamud compatibility.");
             }
         }
+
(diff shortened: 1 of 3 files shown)


XIVLauncher/Windows/SettingsWindow.xaml.cs
```diff
@@ -459,5 +459,10 @@ private void EnableHooksCheckBox_OnChecked(object sender, RoutedEventArgs e)
                 Log.Error(exc, "Couldn't check dalamud compatibility.");
             }
         }
+
+        private void SetupTabControl_SelectionChanged(object sender, SelectionChangedEventArgs e)
```
goaaats: This seems to be left in by accident - I'll revert it after merging

Issue #136, opened 2020-01-01
Not autologin


Mara-Li:
Since the last update, the launcher wipe my informations for connexion and not autologin. 

Happy new year of bug, Goaat ! x)

Dinokin:
can confirm. autologin not working for me.

werewolfboy13:
same issue here

Aida-Enna:
Solved in pull #137, closing. Thank you for reporting!

PR #137, opened 2020-01-01
Fix issue #136

Dinokin:
The issue was caused because no default account was setup in the settings. So i changed the account manager to load the first account in the account list only if the account list has just one item.


goaaats:
Thanks. This should be moved into the json configuration, must have missed it during the refactor.

PR #176, opened 2020-02-29
fix: roulette chan button opening duty finder ui

fmauNeko:
Hello,
This is a fix for the "Set the Roulette Bonus notification channel." opening the Duty Finder notification channel instead.


Aida-Enna:
The roulette bonus notification didn't break, only knowing which content you would get *from* the roulette did.

PR #199, opened 2020-03-30
Fix for when additionalArguments is null

pmgr:
Fix error with parsing additionalArguments when it's null.

### DIFF
--- a/XIVLauncher/Game/XivGame.cs
+++ b/XIVLauncher/Game/XivGame.cs
@@ -206,9 +206,12 @@ namespace XIVLauncher.Game
                 }
 
                 // This is a bit of a hack; ideally additionalArguments would be a dictionary or some KeyValue structure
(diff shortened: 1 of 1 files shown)


XIVLauncher/Game/XivGame.cs
```diff
@@ -206,9 +206,12 @@ public LoginResult Login(string userName, string password, string otp, bool isSt
                 }
 
                 // This is a bit of a hack; ideally additionalArguments would be a dictionary or some KeyValue structure
-                var regex = new Regex(@"\s*(?<key>[^=]+)\s*=\s*(?<value>[^\s]+)\s*", RegexOptions.Compiled);
-                foreach (Match match in regex.Matches(additionalArguments))
-                    argumentBuilder.Append(match.Groups["key"].Value, match.Groups["value"].Value);
+                if (additionalArguments != null && additionalArguments != "")
```
goaaats: !string.IsNullOrEmptry() :p

PR #218, opened 2020-05-18
Add option to disable DF taskbar flashing

Poliwrath:
depends on https://github.com/goatcorp/Dalamud/pull/114


goaaats:
We should probably move away from c# settings completely soon, but this is fine for now. Thanks!

PR #223, opened 2020-06-17
Patch downloading

goaaats:
- [x] Handle boot patches
- [x] Handle ex patches
- [x] Start game after patch done
- [x] More accurate version check
- [x] Correct patch file naming
- [x] Loc
- [x] Testing


goaaats:
Needs to be checked on Linux, but that can be taken care of later

PR #244, opened 2020-07-22
Various changes

Poliwrath:
* Create a separate build configuration that sets XL_NOAUTOUPDATE (I hope I don't break CI somehow)
* Remove x86 / x64 platforms since it's always built as Any CPU
CI log:
```
2020-07-21T17:55:41.5300245Z Project "D:\a\FFXIVQuickLauncher\FFXIVQuickLauncher\XIVLauncher.sln" on node 1 (Build target(s)).
2020-07-21T17:55:41.5323557Z ValidateSolutionConfiguration:
2020-07-21T17:55:41.5323921Z   Building solution configuration "Release|Any CPU".
```
* Move repo URL to a separate const string since the old URL (``https://github.com/goaaats/FFXIVQuickLauncher/``) was still being used in some places
* Modify the default game path checks to check every drive for the game folder (example: ``D:\SquareEnix\FINAL FANTASY XIV - A Realm Reborn``), and also use ``Environment.SpecialFolder.ProgramFilesX86`` instead of hardcoding "Program Files (x86)", since that directory doesn't exist on non-english 32-bit OSes afaik

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Build DotNet4
         run: |
           cd "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\"
-           .\MSBuild.exe $Env:GITHUB_WORKSPACE\XIVLauncher.sln /t:Build /p:Configuration=Release /p:DefineConstants=XL_NOAUTOUPDATE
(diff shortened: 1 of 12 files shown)


goaaats:
x64/x86 were probably a remnant from when Nhaama was still included, thank you!

Poliwrath:
Should be good now, I thought I accidentally added that IWshRuntimeLibrary thing

goaaats:
Thanks, was gonna leave it open otherwise

PR #254, opened 2020-08-11
fix about tab

Poliwrath:


goaaats:
We should get rid of all of the x:Uid at some point since they were for the .NET localization toolkit which sucks, transitioned over to my own a while ago. Thanks!

Poliwrath:
Oh, in that case, I'll reopen this once that's all finished

PR #294, opened 2020-10-29
Add support for setting language as launch param

reiichi001:
Adds a new launch parameter --clientlang=LANG to allow users to override saved settings and language the client in another language as defined in ClientLanguage enum. 

Also handle the --account and new --clientlang params in either order now that a user may want both.

### DIFF
--- a/XIVLauncher/App.xaml.cs
+++ b/XIVLauncher/App.xaml.cs
@@ -196,48 +196,41 @@ namespace XIVLauncher
 
         private void App_OnStartup(object sender, StartupEventArgs e)
         {
-            if (e.Args.Length > 0 && e.Args[0] == "--genLocalizable")
-            {
-                Loc.ExportLocalizable();
-                Environment.Exit(0);
-                return;
-            }
-
-            if (e.Args.Length > 0 && e.Args[0] == "--genIntegrity")
-            {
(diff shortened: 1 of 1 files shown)


XIVLauncher/App.xaml.cs
```diff
@@ -216,10 +216,32 @@ private void App_OnStartup(object sender, StartupEventArgs e)
...
+                {
+                    if (arg.StartsWith("--account="))
+                    {
+                        accountName = arg.Substring(arg.IndexOf("=", StringComparison.InvariantCulture) + 1);
+                        App.Settings.CurrentAccountId = accountName;
+                    }
+                }
+            }
+                
+
```
goaaats, line 231: You can combine these for loops to simplify the logic a bit!

goaaats (changes requested):

XIVLauncher/App.xaml.cs
reiichi001: I've updated the PR to include a simplified loop and dragged in the other two launch args because why not. They'll exit as usual if selected.

I've also tested what happens if a user selects an unsupported language and the enum parser will default to Japanese because it's the first one available. As this will be a command for advanced users, I think this is probably fine since we can make an accompanying FAQ for it if needed.

goaaats:
thank you franz, very nice

PR #298, opened 2020-10-31
feat: added launcher language option

Aireil:
Not sure how I feel about the LauncherLanguage.cs. I used the same idea as the client language, but there were some differences that required it.
There is probably a better way to do this; I'm open to any feedback.

Edit: I added Portuguese as I saw there was already a file for it, and the localization is complete on Crowdin. Russian and Norwegian are also 100%, but don't have a file yet.

### DIFF
--- a/XIVLauncher/App.config
+++ b/XIVLauncher/App.config
@@ -20,6 +20,9 @@
             <setting name="Language" serializeAs="String">
                 <value>1</value>
             </setting>
+            <setting name="LauncherLanguage" serializeAs="String">
+                <value>1</value>
+            </setting>
             <setting name="NeedsOtp" serializeAs="String">
                 <value>False</value>
             </setting>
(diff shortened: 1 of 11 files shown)


XIVLauncher/App.config
```diff
@@ -20,6 +20,9 @@
             <setting name="Language" serializeAs="String">
                 <value>1</value>
             </setting>
+            <setting name="LauncherLanguage" serializeAs="String">
```
goaaats, line 23: We aren't using these settings anymore - you don't have to add them when added to LauncherSettingsV3. They're just here for compatibility.

goaaats (changes requested):

Aireil:
Removed the legacy settings and added a notice concerning the need to restart the launcher.

XIVLauncher_2020-12-06_13-57-39

goaaats (approved):

PR #340, opened 2020-12-21
Replace the check boot version error with a nice message

Aireil:


XIVLauncher/Windows/MainWindow.xaml.cs
```diff
@@ -251,7 +251,22 @@ private void HandleBootCheck(Action whenFinishAction)
...
+                try
+                {
+                    bootPatches = _launcher.CheckBootVersion(App.Settings.GamePath);
+                }
+                catch (Exception ex)
+                {
+                    Log.Error(ex, "Unable to check boot version.");
+                    MessageBox.Show(Loc.Localize("CheckBootVersionError", "XIVLauncher was not able to check the boot version for the select game installation. This can happen if a maintenance is currently in progress or if your connection to the version check server is not available. Please report this error if you are able to login with the official launcher, but not XIVLauncher."), "XIVLauncher", MessageBoxButton.OK, MessageBoxImage.Error);
+
+                    _ = Task.Run(SetupHeadlines);
```
goaaats, line 264: Why the discard?

XIVLauncher/Windows/MainWindow.xaml.cs
```diff
@@ -251,7 +251,22 @@ private void HandleBootCheck(Action whenFinishAction)
...
+                    Show();
+                    Activate();
+                    return;
```
goaaats, line 267: You also need to set isLoggingIn = false here, bit messy, should redo error handling in MainWindow some time

goaaats (changes requested):

XIVLauncher/Windows/MainWindow.xaml.cs
Aireil: Oops, yep, ctrl z more than I thought. In my opinion, it shouldn't need to Show() and Activate(), just return normally and handle it there. I just copied the maintenance error as this is the same thing.

XIVLauncher/Windows/MainWindow.xaml.cs
Aireil: It was just to suppress the warnings in the Task, not needed there though.

goaaats (approved):

Issue #342, opened 2020-12-23
If you enter a bad password and check automatic connexion, FFXIVQuickLauncher does not request a new password


legluondunet:
Hello,
I entered a bad password and checked "automatic connexion".
So FFXIVQuickLauncher quit and now when I launch it only shows me this error message:

![ArcoLinux_2020-12-23_12-35-10](https://user-images.githubusercontent.com/325405/102992764-22317700-451c-11eb-92f7-6a2c78a86bb5.png)

I expected it ask me to enter a new password.

I had to manually delete json files from /home/legluondunet/Games/final-fantasy-xiv-a-realm-reborn.bak/drive_c/users/legluondunet/Application Data/XIVLauncher/ folder, it is not very user friendly.

Zyian:
You can hold **shift** when you open the Launcher in order to disable the autolaunch. The simple fix for this would be to ensure you have entered the correct password before you enable autolaunch

legluondunet:
"The simple fix for this would be to ensure you have entered the correct password before you enable autolaunch"
I entered the correct password but today Square Enix disabled my account to force me to change password. 
Anyway it would be more user friendly to just return to the interface after clicking on "ok" on this error dialog box.
Close all windows after this error dialog box is a little "brutal".

Aida-Enna:
Solved in #348.

PR #348, opened 2021-01-05
Disable Autologin when OAuth fails

Zyian:
This is a quick and dirty change to disable the auto login when OAuth fails, the issues that come up from OAuth failing are numerous, but the most common case is that the credentials were typed in wrong. Since many people are not aware of the holding Shift to disable auto login, adding it here to the failure should serve the purpose well.

#342


Zyian:
Changed the MessageBox here to give the OKCancel buttons if autolaunch is enabled. OK will disable autolaunch for the next run, Cancel will retain original behavior. I like this more than the other way of just always disabling the feature.

goaaats (approved):

goaaats:
I'll need to localize this after merging

PR #364, opened 2021-01-26
Improvements to patch downloader, restrict parallel downloads

### LINKED ISSUE #333: Choose the number of parallel downloads
**What is your suggestion/feature request?**
Right now XIVLauncher downloads 4 patches concurrently. I'd like to be able to change how many concurrent downloads are allowed, as 4 concurrent downloads is a bit too much for my connection and makes it very unstable, sometimes causing the patcher to fail the download.

**Describe how you'd like this to work**
I'd like an extra setting in the launcher's settings that lets you choose a number of concurrent downloads (like something between 1 and 6) with the default setting on 4.

Zyian:
Starting this early for feedback

- Don't think I would need to shorten the size of the array, but it probably would be more polished
- The UI component is basically the same as download speeds.

This works to resolve #333

### DIFF
--- a/XIVLauncher/Game/Patch/PatchManager.cs
+++ b/XIVLauncher/Game/Patch/PatchManager.cs
@@ -51,7 +51,7 @@ namespace XIVLauncher.Game.Patch
                 UserAgent = "FFXIV PATCH CLIENT",
                 Accept = "*/*"
             },
-            MaximumBytesPerSecond = App.Settings.SpeedLimitBytes / MAX_DOWNLOADS_AT_ONCE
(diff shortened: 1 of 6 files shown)


Zyian:
Functionally this works well, depending on how many parallel downloads you set (default 4) it will use the download limit accordingly, 1 thread at max speed or more threads split evenly.

There could be some more polish on the UI side, such as hiding the other download bars when not in use.

If there's anything I should add or change, I'm open to it

peter7north (approved):

Zyian:
This is so old I'm just gonna close this and come back to it at another point in time.

PR #366, opened 2021-01-29
Display time left when GitHub rate limit exceeded

Aireil:
Not sure if useful, but it avoids the confusion of not knowing when it will be available again.

Message


goaaats:
I wish we wouldn't need this

Aireil:
To be fair, you can open the launcher 60 times every hour, so not sure how you even run out 🤔 . Using GitHub API for other things I guess.

LysitheaStar:
Could probably as well add a "launch anyway" button or something like that

PR #374, opened 2021-02-06
Don't start patcher in admin mode if user has write access to folder 

jcb936:
Self-explanatory, useful if trying to install FFXIV on a computer where you don't have admin rights

### DIFF
--- a/XIVLauncher/Game/Patch/PatchInstaller.cs
+++ b/XIVLauncher/Game/Patch/PatchInstaller.cs
@@ -37,7 +37,7 @@ namespace XIVLauncher.Game.Patch
 
         public InstallerState State { get; private set; } = InstallerState.NotStarted;
 
-        public void StartIfNeeded()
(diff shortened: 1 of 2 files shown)


jcb936:
Bug with implementation, will open again in a bit

goaaats:
I actually think that it'd be better to just have this as an env var - since, when you change the game path, you'd need to relaunch the patcher if it's now in a location that requires elevation to access. Let's keep complexity down, and make this a Power-User feature, I guess.

jcb936:
> 
> 
> I actually think that it'd be better to just have this as an env var - since, when you change the game path, you'd need to relaunch the patcher if it's now in a location that requires elevation to access. Let's keep complexity down, and make this a Power-User feature, I guess.

That's alright, I have now implemented it using environment variables. Needed to open a new PR tho because git messed up a bit, you can find it here:
#376

Issue #380, opened 2021-02-10
Launcher using different drive for temporary files resulting in out of space errors without prior warning


Ihysoal:
**Describe the bug here:**
It seems that the launcher by default is downloading all the patch files to AppData (and sporadically cleans some of it up?) without regard for space.
This resulted in multiple app crashes as it kept hitting the limit of the 20GB free space on it when I had a perfectly fine installation folder with 160 GB of space available and where it's installing to. It also just causes unnecessary writes on another SSD, not that I'm concerned about this killing lifespan.

As I was writing this bug report and restarting for the fourth time the application finally notified me that the disk was full and I could change the patch folder, something I hadn't noticed myself. I'm not sure if the option presented itself during setup or that I missed it. In case I missed it then I'm a dummy. However, it still crashed on me instead of stopping me more gracefully.

**Steps to reproduce the bug:**
- Have not enough disk space on the drive that the user's AppData folder is on, typically C:
- Install to another disk (in my case: F:\Games\FINAL FANTASY XIV - A Realm Reborn)
- No warnings about disk space
- Application crashes when it inevitably runs out of space to write
- Practice your ad pop-up closing skills by closing a multitude of error prompts.

**Expected behavior**
- Warn the user about not enough download folder space.
- Download to the installation folder by default or a specified download folder during installation.

**Other information (*please* complete the following information):**
  Operating System: Windows 10 Pro
  XIVLauncher version: 5.2.4
  Launching as admin?: No
  Using Steam?: Nope
  Addons/Plugins enabled: Ixnay

**Additional information**
Also, it seems to download to AppData/Roaming by default which is _technically_ a big no-no as that folder is meant for important configuration data that, well, roams between multiple PCs following the user's account on a domain. Although I doubt many people will be installing it in such an environment...

Edit: Small theory: issue only cropped up because download speed exceeded installation/unpacking speed.

reiichi001:
You can change the path that patch downloads are saved to in the xivlauncher settings. 

It also sounds like you did not set your game path correctly if XIVLauncher is trying to fully reinstall the client. I've copied our FAQ from Discord on this topic, as it will likely help.

**Where can I find my FFXIV Installation?**
(AKA: What does XIVLauncher mean by GamePath?)

FFXIV installs to a few different locations depending on whether you used the official installer or steam, when you installed it, and potentionally if you installed the free trial or not. Here are some of the common paths.

[...]

PR #381, opened 2021-02-10
Support for patch download checks

reiichi001:
Adds support for some additional patch download checks, even on debug builds.

Downloader library updated to 1.7.0 which is new enough to support their temp directory override, but not have breaking changes.

Added a disk space check that will check for enough space for ALL patches destined for the patch folder, because we can't guarantee they'll download in order. (And often on a base game install, they absolutely won't).

Should make the user in https://github.com/goatcorp/FFXIVQuickLauncher/issues/380 happy too.


Ihysoal:
Thank you for the effort, and sorry for the vague issue.

goaaats (approved):

Issue #407, opened 2021-04-04
Reproducible builds


goaaats:
We should automate builds and releases to prevent problems and lessen workload.

Artifacts are already generated on build.

goaaats:
We have this working nicely now, thanks Zyian

PR #425, opened 2021-04-27
Modify ci-workflow.yml to automate creating releases

Zyian:
This automates the process of creating new releases for XIVLauncher, currently only triggers on tags, and the description of the release will need to get updated after the action is complete.

Example of a completed run:
image


Zyian:
This will also resolve #407

goaaats (approved):

PR #432, opened 2021-05-20
Added option to change Refresh Rate when Launching Game

trowgundam:
First, let me say this change comes out of a weird issue I have running the game. For whatever reason, my game will inevitably crash pretty frequently. I finally tracked down the issue to me running at refresh rates higher than 60, not even capping my FPS at 60 stops it. Well, I've gotten fed up manually changing my refresh rate before launching the game and, inevitably, forgetting to change it back after I quit the game.

I had been writing my own tool to change the refresh rate when detecting specific applications launching, but since this is the only game I have such an issue, that seemed like overkill (plus I'm a bit lazy). However, adding an option to this project is considerably easier. What I have done is as follows:

1. Added another setting, located directly underneath the DirectX option, to specify a Refresh Rate. A value of 0, which should be the default, should result in the same behavior from before this change.
2. Added logic to the launch code to change the refresh rate before starting the application, and if successful, save the refresh rate before the change, so it can be restored.
3. On the game exiting, it will attempt to restore the saved refresh rate.

As you can see nothing terribly drastic. I do have a few points of concern or changes that might be called for. First, I don't know other languages, however the text I have added is rather simple and I'd bet Google Translate could help. However, I wasn't certain what all I need to do in order to properly add the Localization for my items. Second, I am not using any obscure functions from the Windows API, and as a result I don't think this would have a problem running under Wine in Linux, but I am unable to test this. I wanted to just add a check for Wine and disable the logic, but wasn't certain how. Finally, while nothing I've done will crash, there is nothing to validate that the user picks a valid refresh rate. If an invalid refresh rate is requested to the Windows API will just return a failure value stating a bad parameter was provided. I wanted to add a validation, which is fairly easy to do, but I did not see an immediate example to base my own validation off of.

If there are any suggestions or changes you want, please let me know, and I will attempt to make them.


goaaats:
Hello! Thank you for your contribution!
Your code looks good, absolutely no objections there - I do however think that adding this complexity to XIVLauncher for a specific edge case like this might be a bit overkill.

Just to make sure, setting your FPS to 30 something in the game doesn't work? Have you excluded causes like driver issues?

We've not ever heard of this specific issue before, which is why I am curious since that is usually the telling factor for what to implement.

trowgundam:
I've tried many things for my issue. I've been dealing with since at least Shadowbringer's launch. I've found several things that make my crash less likely to occur, but it still does occur. For whatever reason the DCH version of the Nvidia driver crashes more often. Limiting my framerate does make the crash occur less often, but it still does occur. So far I've found two things that stop my crashing entirely. First is using the Nvidia driver that was released about a month before Shadowbringers. I would stick to that, but there have been numerous critical security flaws patched since then, and that isn't advisable anymore. That was actually my solution for a while. The second is just setting my Refresh Rate to 60. I'm not sure why doing this versus just limiting my FPS to 60 react differently.

I'm certain this is some issue specific to my computer. I'm also almost certain it is something up with my GPU. I know for a fact that the same Nvidia driver works just fine on at least 2 other computers, even at higher refresh rates. I also know multiple others with a RTX 2080 Ti that have absolutely no issues, even with higher refresh rates. The issue has also persisted through multiple installs of Windows 10, and even happens (albeit much rarer) when running under Linux. The game has never really like my GPU, and I'm hoping to get a new one when the 3080 Ti launches, but well I'm sure you are aware of the GPU Market right now. So that will likely be a fruitless endeavor.

If this complexity seems too much, I get that. I'd accept that, but what about something a bit different. The launcher has the ability to start programs when you start the game already, how about the opposite, i.e. Run/Start programs when the game is exited? That way I could run a command to change my Refresh Rate when the game launches, and then auto run the command to change it back afterwards. It'd also be useful so I could run my script for changing the fan speed on my Commander Pro (I prefer not to use iCUE and instead manually control them through liquidctl).

trowgundam:
I have implemented the alternative I mentioned at the end of my last comment with Pull Request #438. If that is a more acceptable method, feel free to deny and close this Pull Request.

trowgundam:
No point keeping this out here anymore. The branch is out of date and ultimately not needed anymore.

PR #436, opened 2021-05-30
fix: use correct release mode depending on build

Zyian:
This will build the launcher in the correct Configuration so that:
- master branch: ReleaseNoUpdate
- tags: Release

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -2,6 +2,60 @@ name: Build XIVLauncher
 on: [push, pull_request]
 
 jobs:
+  build-release:
+    name: Build Relase on Windows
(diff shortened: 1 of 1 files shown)


Zyian:
Updated workflow to make a little more sense, I split the Release logic into its own build vs the regular artifact uploading, rather than checking for tags per step, I just checked on whether the job should run at all, keeps it cleaner and more maintainable.

Examples:
- Tags: https://github.com/Zyian/FFXIVQuickLauncher/actions/runs/900272954
- Master: https://github.com/Zyian/FFXIVQuickLauncher/actions/runs/900271939

Also pulls in latest change to mark releases initially as pre-releases

goaaats (approved):

PR #438, opened 2021-05-30
Added option for Auto Launch stuff to run on game close

trowgundam:
Added option to only run an Auto-Launch program on Game Close. Not much to the change other than that. This provides a far more general way to solve the problem I was trying to solve with the change from #432 .


goaaats (approved):

goaaats:
All good, mistake from an earlier commit

PR #479, opened 2021-07-11
Build customized aria2c

reiichi001:
This will run correctly on Windows and Linux. Here's the configure I used to compile using msys2 mingw64 x86_64-w64-mingw32

# aria2c-openssl_libz
./configure \
    --with-included-gettext \
    --disable-nls \
    --without-libcares \
    --without-gnutls \
    --without-wintls \
    --with-openssl \
    --without-sqlite3 \
    --without-libxml2 \
    --without-libexpat \
    --with-libz \
    --without-libgmp \
    --without-libssh2 \
    --without-libgcrypt \
    --without-libnettle \
    ARIA2_STATIC=yes


reiichi001:
This is no longer needed. Wine 6.15 and later redid winsock and the standard aria2c binary works now.

Issue #489, opened 2021-07-20
Add NA Lodestone News banners to Launcher


Zyian:
Currently when XL is set to the English client language it will only load Lodestone news banners from the EU lodestone. We should add another language to separate out EU and NA lodestone news. Couple of ways we could do this:

- Add a new client language that maps to English when launching the game but loads NA news (`English - NA` and `English - EU`)
- Or decouple Client Language from Lodestone News and have a separate setting for this to select from.

Location in question: https://github.com/goatcorp/FFXIVQuickLauncher/blob/f5459dd30191ad71ab98038a0f04f4f9d7116853/src/XIVLauncher/Windows/MainWindow.xaml.cs#L854

PR #490, opened 2021-07-20
feat: show NA news when region is NA (#489)

karashiiro:
This uses the Globalization API, rather than add a new client language or an option.


karashiiro:
Whoops, committed an old submodule file, will force-push in a minute to remove that.

karashiiro:
Corrected.

goaaats (approved):

PR #516, opened 2021-08-21
fix: incorrect warnings with other d3d11 and dxgi exist

nopjmp:
We do not want to warn if someone is using DXVK or another API wrapper.

### DIFF
--- a/src/XIVLauncher/Game/ProblemCheck.cs
+++ b/src/XIVLauncher/Game/ProblemCheck.cs
@@ -89,17 +89,23 @@ namespace XIVLauncher.Game
 
             if (d3d11.Exists && dxgi.Exists)
             {
-                if (MessageBox.Show(
-                    Loc.Localize("GShadeError", "A broken GShade installation was detected.\n\nThe game cannot start. Do you want XIVLauncher to fix this? You will need to reinstall GShade."),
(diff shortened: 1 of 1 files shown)


src/XIVLauncher/Game/ProblemCheck.cs
```diff
@@ -89,17 +89,23 @@ public static void RunCheck()
...
-                    }
-                    catch (Exception ex)
+                    if (MessageBox.Show(
+                        Loc.Localize("GShadeError", "A broken GShade installation was detected.\n\nThe game cannot start. Do you want XIVLauncher to fix this? You will need to reinstall GShade."),
+                        "XIVLauncher Error", MessageBoxButton.YesNo, MessageBoxImage.Error) == MessageBoxResult.Yes)
                     {
-                        Log.Error(ex, "Could not delete duplicate GShade.");
+                        try
+                        {
+                            dxgi.Delete();
```
reiichi001, line 103: We shouldn't try to delete it. Just make an error message. A user could have specific reasons for using a particular injector.

goaaats: No one should use two GShades and I don't want to relocalize this

nopjmp: This was pre-existing code and behavior. It doesn't seem to work on Windows at least, but might work on Wine I haven't tested it yet.

PR #572, opened 2021-10-30
enable authentication as mac client via `XL_AUTHENTICATEASMAC`

marzent:
This is needed for people who only own a Mac license but still want to launch their game via xivlauncher.


goaaats:
Hey, thanks for your PR.

We aren't implementing this on purpose since we don't want to circumvent SE's platform restrictions, as arbitrary as they might be. Additionally, I have no way to verify/test the login flow on Mac for myself at the moment.

We're working on a multiplatform XIVLauncher rewrite right now that will support Mac OS and Mac licenses - it's a fair bit of work still though.

marzent:
Right now xivlauncher is effectively circumventing SE platform restrictions on mac anyways by allowing Windows licenses to login on Mac. 
The official launchers behavior is to check for the`wine-get-version` export inside `ntdll.dll` and if it exists treats it as Mac version. 
Would a similar check be a suitable workaround since this will be very hard to fake on Windows and would give xivlauncher the exact same authentication behavior as the native launcher .exe which is identical on all platforms?

goaaats:
I'll look into it - will set up a Mac VM to do some captures for myself soon, likely after endwalker release.

marzent:
> Right now xivlauncher is effectively circumventing SE platform restrictions on mac anyways by allowing Windows licenses to login on Mac.
> The official launchers behavior is to check for the`wine_get_version` export inside `ntdll.dll` and if it exists treats it as Mac version.
> Would a similar check be a suitable workaround since this will be very hard to fake on Windows and would give xivlauncher the exact same authentication behavior as the native launcher .exe which is identical on all platforms?

Possible combinations are:
1) Windows Authentication -> Windows Client
2) Windows Authentication -> Mac Client (xivlauncher currently does this)
3) Mac Authentication -> Mac Client
4) Mac Authentication -> Windows client (possible with the original version of this commit on Windows)

Implementing the WineExports check would remove options 2) and 4) and prob make xivlauncher safer to use as well since I guess 2) & 4) look very suspicious on SE side

marzent:
I've updated the PR to only allow 1) and 3).
This has the added benefit that if someone forgot to set `HKEY_CURRENT_USER\Software\Wine\HideWineExports` on Linux they will see a *Wine on Mac* window title reminding that a Mac Authentication -> Mac Client will take place if they proceed.

marzent:
> I have no way to verify/test the login flow on Mac for myself at the moment.

I also tested and verified this on Linux, the Mac Client will have different Graphics and Keybind options

marzent:
I've marked this as draft since there seems to be issues with patching atm

marzent:
Every possible configuration I could think of is working now atm, just make sure that if you do testing that it is done with an account which *only* has a Mac license, since SE doesn't seem to care at all about the useragent at all for accounts that have both licenses activated.

marzent:
> I'll look into it - will set up a Mac VM to do some captures for myself soon, likely after endwalker release.

can send you captures on any platform with any type of account (native or xivlauncher) if that would help

src/XIVLauncher/Settings/EnvironmentSettings.cs
```diff
@@ -1,17 +1,25 @@
...
 using System.Threading.Tasks;
 
 namespace XIVLauncher
 {
     internal static class EnvironmentSettings
     {
-        public static bool IsWine => CheckEnvBool("XL_WINEONLINUX");
+        public static bool IsWine => IsLinux || IsMac;
+        public static bool IsLinux => CheckEnvBool("XL_WINEONLINUX");
+        public static bool IsMac => GetProcAddress(GetModuleHandle("ntdll.dll"), "wine_get_version") != IntPtr.Zero;
```
reiichi001, line 14: Rather than set the `IsMac` bool here like this, I think we'd benefit from a `XL_WINEONMAC` environment variable. Then we can adjust `IsWine` accordingly because it's also possible for Linux users to turn off Wine exports.

The current `IsMac` check would probably best be turned into something like `FoundWineExports` and we could always do something like this to handle all possible cases.
```
IsWine = IsLinux || IsMac || FoundWineExports;
```

This is mostly to handle any potential future cases where we need to make platform-specific changes. There are quite a few for Linux and we can't guarantee that Wine on Linux and CrossOver on macOS will always be perfectly in sync in terms of experience, features, or bugs.

marzent: Should be added in with 42196e2aeeb8e638be2a7b9ba3ae0591ae3f9bde

goaaats:
Might have to rebase this again, sorry, but I'll have a look soon. Thanks for waiting.

marzent:
> Might have to rebase this again, sorry, but I'll have a look soon. Thanks for waiting.

No worries, maybe it would make sense to add an UI element that indicates which authentication is taking place (since `IsMac` no longer has anything to do with the login)?

Could be grouped with the steam service account checkbox somehow

lambroisie:
I can confirm this works on Linux with a patched Wine version.

PR #595, opened 2021-12-02
Fix: wait for bootPatches

KattTails:
Seems like bootPatches isn't initialised before being read later?


goaaats:
Thanks, I did a big emergency refactor of this a little while ago, must have overlooked it.

PR #697, opened 2022-01-22
Do not exit when rate limited

ryknuth:
I am behind a double or triple NAT. GitHub's rate limiting is by IP. There is someone on my IP that immediately blows through the GitHub rate limit. It's frustrating as there is nothing I can do about it. XIVLauncher exits if it can't check for an update. I have changed this to allow the app to continue if the exception was caused from rate limiting.

### DIFF
--- a/src/XIVLauncher/Updates.cs
+++ b/src/XIVLauncher/Updates.cs
@@ -60,7 +60,14 @@ namespace XIVLauncher
                                 "XIVLauncher",
                                  MessageBoxButton.OK,
                                  MessageBoxImage.Error);
-                System.Environment.Exit(1);
+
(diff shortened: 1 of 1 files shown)


goaaats:
Hi, this is deliberate at the moment, since troubleshooting for users on multiple XL versions is very cumbersome for us and there might be changes we need to do to the launcher that should reach all users immediately, e.g. when SE changes the way they communicate with the login server.

You can set the variable `XL_NOAUTOUPDATE` to `true` to disable updates, but you should be aware of what this entails. Further down the road we are planning to switch to a CDN instead of using the github api.

ryknuth:
I tried out the prerelease and it worked. Thank you.

PR #718, opened 2022-02-21
Attempt to fix multithreading problems

Soreepeong:
* Use Array.SyncRoot instead of array instance itself
* Create FileStream per thread instead of sharing it with a lock


Soreepeong:
Nah this still didn't fix the problem.

PR #742, opened 2022-03-08
PatchVerifier changes and ensure defaults for TreatNonZeroExitCodesAsFailure

Soreepeong:


Soreepeong:
* PatchVerifier will now skip verification if corresponding expac is not purchased by the currently logged in user.
* PatchVerifier will now log more, to figure out why are some .patch files missing.
* TreatNonZeroExitCodeAsFailure has multiple references with scattered around default values, and this normalizes defaults to false. Default value should be ensured somewhere but this is how it is for now.

goaaats (approved):

PR #744, opened 2022-03-09
Preliminary work for new Blowfish implementation

Minoost:
Historically the blowfish implementation appears in Square Enix games
has (had) been riddled with numerous bugs (e.g. misinterpreting
a unsigned integer into signed one)

This one was one of them where you are suppose to interpret 8 bytes into
two `u32` in big-endian order in the process of constructing fesitel cipher
but they instead interpreted them (or rather, what's more likely, they simply didn't bother)
as two `u32` in **little-endian** order.
(though this isn't strictly necessary but by conventions so that you could
encrypt a plaintext on program A and decrypt them on program B and
it would still gives you the same data back)

With the recent launcher update on March 7, they switched to more *sane*
implementation for Steam authentication and this time actually uses BE in their code.

This means we can actually grab `bouncycastle` package from nuget (or just any library with correct impl)
and it will remain compatible from now on!

But as we're pretty sure that authenticating with SQEX uses old Blowfish code
while new steam stuff is using new Blowfish path, and because we want to keep dependencies small,
I actually ported my blowfish implementation from `Aither.Crypto`, a blowfish implementation that matches both the old and new behavior.

### DIFF
--- a/src/XIVLauncher.Common/Encryption/BlockCipher/Blowfish.cs
+++ b/src/XIVLauncher.Common/Encryption/BlockCipher/Blowfish.cs
@@ -1,0 +1,93 @@
+// NOTE: This file is copy-pasted almost *as-is* from the previous work `Aither.Crypto`
+//       hence currently it does not follow XL's naming convetions.
+//       
+//       It's totally okay to change this. But for now, this is what it is atm.
+// ReSharper disable InconsistentNaming
+
+using System;
+using System.Buffers.Binary;
+
+namespace XIVLauncher.Common.Encryption.BlockCipher
+{
+    public sealed class Blowfish : IBlockCipher
+    {
+        /// <inheritdoc />
+        public int BlockSize => 8;
+
+        // NOTE: this field should never be marked as readonly as it actually creates a defensive copy on every access. (it's a trap)
+        // https://devblogs.microsoft.com/premier-developer/the-in-modifier-and-the-readonly-structs-in-c/
+
+        private BlowfishState m_state;
+
+        /// <summary>
+        /// Initializes a new instance of the Blowfish class.
+        /// </summary>
+        /// <remarks>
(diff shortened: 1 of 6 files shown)


src/XIVLauncher.Common/Encryption/BlockCipher/Blowfish.cs
```diff
@@ -0,0 +1,93 @@
...
+            if (BitConverter.IsLittleEndian)
+            {
+                xl = BinaryPrimitives.ReverseEndianness(xl);
+                xr = BinaryPrimitives.ReverseEndianness(xr);
+            }
+
+            (xl, xr) = m_state.EncryptBlock(xl, xr);
+
+            // will be elided by JIT
+            if (BitConverter.IsLittleEndian)
```
Minoost, line 53: just a note: you can change these two lines (line 44 and 53) if you have to switch back to old (little-endian) blowfish behavior.

Minoost:
Apparently .NET standard 2.1 isn't supported on .NET framework at all. 👀  I guess we could probably use bouncycastle meanwhile.

https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1#net-standard-versions

goaaats:
What do you need from 2.1? We definitely need to use 2.0, otherwise XL can't use the library

wolfcomp:
From what I can read from the code, it doesn't use any 2.1 features, so 2.0 should work.

Minoost:
> What do you need from 2.1? We definitely need to use 2.0, otherwise XL can't use the library

> From what I can read from the code, it doesn't use any 2.1 features, so 2.0 should work.

https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=net-6.0#applies-to

`Span<T>` actually is. Albeit you could manually pull `System.Memory` from nuget on anything <.netstd 2.0, it will be slightly less efficient. (I guess this is more than enough™️ for now)

goaaats (approved):

PR #780, opened 2022-03-21
Accept --steamticket command line parameter

Jadd:
Allows user to provide --steamticket=***** command line parameter as a base64 encoded string.

I am working on an open-source alternative to using steam using this third-party library: https://github.com/DoctorMcKay/node-steam-user/

The idea would be to obtain the steam auth session ticket and pass it to XIVLauncher to do the heavy lifting. This would be beneficial to any user with issues running steam, or those attempting to launch multiple clients without logging in and out of steam.


goaaats:
If you want, you can include a sample script in the misc folder.

PR #785, opened 2022-03-27
Upstream Squirrel => Clowd.Squirrel

Dormanil:
## Introduction

Old Squirrel was clunky and doesn't really support modern .NET. Also, recent wine-GE builds just don't install with old Squirrel. 

## Overview of changes in this PR

This PR will:  
- Remove the unused AssemblyInfo.cs
- Add a Assembly Manifest to indicate Squirrel Awareness (and lack of Windows 7 Compatibility)
- Update, fix warnings (and add nullability check because why not) for XIVLauncher/Updates.cs, including:
  - Adding unused `IAppTool` variable `t`, because the old overload has been obsoleted  
  - Pragma warning disable of the 'Please, PLEASE use ConfigureAwait()' warning, because adding `ConfigureAwait(true)` would be dumb
- Allow downloading of both Setup.exe and XIVLauncherSetup.exe during release creation, because Clowd.Squirrel will name it the latter way and I didn't bother renaming the file since I think that is actually a better name than just Setup.exe
- Create the release using Clowd.Squirrel, and with a pack folder that creates an updated .nuspec for goatsoft instead of goaaats

## Benefits of using Clowd.Squirrel

Changing to Clowd.Squirrel will give us the following benefits:
- Updater is a self-contained .NET 6 executable, and thus
- Allows installation of necessary .NET Framework versions
- Installs necessary Visual C++ Redistributable
- Runs on modern wine and doesn't break on Wine 7.1

## Breaking Changes

- Lutris Installer Script will need removal of the `/silent` argument, because installation of vcredist in silent mode will be cancelled
- In the same vein (but not really breaking), we can remove the vcredist installation in the Lutris Installer Script

## How to test

To facilitate testing, I have created three releases on the [my fork](https://github.com/dormanil/ffxivquicklauncher).  
Release 6.2.210 is identical to the 6.2.21 release here on goatcorp, apart from the change in update server.  
Release 6.2.211 has the upgrade to Clowd.Squirrel. Note that the size of the installer and the delta size is significantly bigger because of the size increase of the updater, and the lack of proper delta support when switching from squirrel.windows to Clowd.Squirrel.  
(Pre-)Release 6.2.215 is apart from the version difference identical to 6.2.211 and there for testing the Clowd.Squirrel -> Clowd.Squirrel update.

### DIFF
--- a/src/XIVLauncher/Updates.cs
+++ b/src/XIVLauncher/Updates.cs
@@ -7,11 +7,13 @@ using Serilog;
 using Squirrel;
 using XIVLauncher.Windows;
 
+#nullable enable
+
 namespace XIVLauncher
 {
     class Updates
     {
-        public event Action<bool> OnUpdateCheckFinished;
+        public event Action<bool>? OnUpdateCheckFinished;
 
         public async Task Run(bool downloadPrerelease, ChangelogWindow changelogWindow)
         {
@@ -26,18 +28,20 @@ namespace XIVLauncher
 
             try
             {
-                ReleaseEntry newRelease = null;
+                ReleaseEntry? newRelease = null;
 
                 using (var updateManager = new UpdateManager(url, "XIVLauncher"))
                 {
                     // TODO: is this allowed?
                     SquirrelAwareApp.HandleEvents(
-                        onInitialInstall: v => updateManager.CreateShortcutForThisExe(),
-                        onAppUpdate: v => updateManager.CreateShortcutForThisExe(),
-                        onAppUninstall: v => updateManager.RemoveShortcutForThisExe());
+                        onInitialInstall: (v, t) => updateManager.CreateShortcutForThisExe(),
+                        onAppUpdate: (v, t) => updateManager.CreateShortcutForThisExe(),
+                        onAppUninstall: (v, t) => updateManager.RemoveShortcutForThisExe());
 
+#pragma warning disable CA2007 // Consider calling ConfigureAwait on the awaited task
                     var a = await updateManager.CheckForUpdate();
                     newRelease = await updateManager.UpdateApp();
+#pragma warning restore CA2007 // Consider calling ConfigureAwait on the awaited task
                 }
 
                 if (newRelease != null)
@@ -89,3 +93,5 @@ namespace XIVLauncher
         }
     }
 }
+
+#nullable disable
--- a/src/XIVLauncher/Updates.cs
+++ b/src/XIVLauncher/Updates.cs
@@ -5,7 +5,9 @@ using System.Windows;
 using CheapLoc;
 using Serilog;
 using Squirrel;
+using Squirrel.SimpleSplat;
 using XIVLauncher.Windows;
+using ILogger = Squirrel.SimpleSplat.ILogger;
 
 #nullable enable
 
@@ -13,7 +15,17 @@ namespace XIVLauncher
 {
     class Updates
     {
-        public event Action<bool>? OnUpdateCheckFinished;
+        public event Action<bool> OnUpdateCheckFinished;
+
+        private readonly SquirrelLogger squirrelLogger;
+
+        public Updates(Action<bool> updateCheckFinished) 
+        { 
+            OnUpdateCheckFinished += updateCheckFinished;
+
+            squirrelLogger = new SquirrelLogger();
+            SquirrelLocator.CurrentMutable.Register(() => squirrelLogger, typeof(ILogger));
+        }
 
         public async Task Run(bool downloadPrerelease, ChangelogWindow changelogWindow)
         {
@@ -30,19 +42,12 @@ namespace XIVLauncher
             {
                 ReleaseEntry? newRelease = null;
 
-                using (var updateManager = new UpdateManager(url, "XIVLauncher"))
-                {
-                    // TODO: is this allowed?
-                    SquirrelAwareApp.HandleEvents(
-                        onInitialInstall: (v, t) => updateManager.CreateShortcutForThisExe(),
-                        onAppUpdate: (v, t) => updateManager.CreateShortcutForThisExe(),
-                        onAppUninstall: (v, t) => updateManager.RemoveShortcutForThisExe());
+                using var updateManager = new UpdateManager(url, "XIVLauncher");
 
 #pragma warning disable CA2007 // Consider calling ConfigureAwait on the awaited task
-                    var a = await updateManager.CheckForUpdate();
-                    newRelease = await updateManager.UpdateApp();
+                var a = await updateManager.CheckForUpdate();
+                newRelease = await updateManager.UpdateApp();
 #pragma warning restore CA2007 // Consider calling ConfigureAwait on the awaited task
-                }
 
                 if (newRelease != null)
                 {
@@ -65,7 +70,7 @@ namespace XIVLauncher
                             };
                         });
 
-                        OnUpdateCheckFinished?.Invoke(false);
+                        OnUpdateCheckFinished(false);
                     }
                     catch (Exception ex)
                     {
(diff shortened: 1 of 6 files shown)


src/XIVLauncher/Updates.cs
```diff
@@ -26,18 +28,20 @@ public async Task Run(bool downloadPrerelease, ChangelogWindow changelogWindow)
 
             try
             {
-                ReleaseEntry newRelease = null;
+                ReleaseEntry? newRelease = null;
 
                 using (var updateManager = new UpdateManager(url, "XIVLauncher"))
                 {
                     // TODO: is this allowed?
                     SquirrelAwareApp.HandleEvents(
```
goaaats, line 36: Should we still be doing this? The Clowd.Squirrel readme recommends using app tool methods instead and doing it when the app starts, not here. updateManager might also be disposed at this point.

goaaats (changes requested):

goaaats:
We should also add a logging sink for Squirrel while we're at this. Here's osu's switchover from earlier this year for reference: https://github.com/ppy/osu/pull/17046/files

src/XIVLauncher/Updates.cs
Dormanil: I moved the handling of Squirrel events to the App startup using tool methods. Squirrel log has also been added to the XIVLauncher log.

goaaats:
Looks fine to me now, did you test fresh installations on a VM? Is the uninstaller entry being created? Start menu folder?

Dormanil:
I should look at my mails more often...

Anyways, I tested installation on a Windows VM; uninstaller entries are created and found in both classic and modern app lists, and the start menu entry is also there.

proof

Dormanil:
@reiichi001, upon merge and official release, the Lutris install script will need the following changes:  
- The task installing the launcher will need to have its argument to install silently removed, or else the installer will cancel when asking to install Visual C++ Redistributable 2015-2022.
- The vcrun2019 winetricks verb should either: 
  - Stay, so that the wine dll overrides are properly in place, but we do download and install vcrun2019 unnecessarily then.
  - Be skipped, instead manually setting `native,builtin` dll overrides to the vcrun2022 dlls.  
  In that case, the following overrides are set by winetricks when installing vcrun2019:
   ```
   *api-ms-win-crt-conio-l1-1-0
   *api-ms-win-crt-heap-l1-1-0
   *api-ms-win-crt-locale-l1-1-0
   *api-ms-win-crt-math-l1-1-0
   *api-ms-win-crt-private-l1-1-0
   *api-ms-win-crt-runtime-l1-1-0
   *api-ms-win-crt-stdio-l1-1-0
   *api-ms-win-crt-time-l1-1-0
   *atl140
   *concrt140
   *msvcp140
   *msvcp140_1
   *msvcp140_2
   *msvcr140
   *vcomp140
   *vcruntime140
   *vcruntime140_1
   ```

reiichi001:
I think we're best keeping winetricks so that overrides are appropriately set. 

Our setup.exe shouldn't be force-installing something if it's already there, right? So since vcrun2019 would already cover what we need, I'm assuming an install check would pass and it wouldn't try anything weird.

Dormanil:
It won't install anything without the user's permission. 

If there's a missing dependency (and vcrun2022 has a different version number than vcrun2019, so vcrun2022 would be missing on a machine with vcrun2019), it asks the user for permission. If it doesn't get permission, it won't install either the dependency or the actual software. In a silent installation, permission is by default not given to anything other than the actual software, and thus a missing dependency results in no installation at all. 

I can change the dependency to vcrun2019 in the squirrel setup, but a silent install will still fail outright should squirrel ever think that the dependency isn't met.

reiichi001:
Winetricks pulls from <https://aka.ms/vs/16/release/vc_redist.x64.exe> too. While the verb is `vcrun2019`, it's the same 2015-2022 from Microsoft, which was formerly 2015-2019.

So under that assumption, users should be getting the same packages as Windows users, with the caveat that wine doesn't run Windows update, so it could potentially require a user to run vcruntime installers again if the user has an old prefix. (But I don't think that'll really be something we need to worry about).

Do you know how this version of SquirrelSetup checks runtimes?

The silent option in Lutris was more of a workaround for permission issues as Wine tries to run everything as the Administrator user otherwise, which would cause the current version of SquirrelSetup we're using to explode and fail on the spot.

Dormanil:
Squirrel pulls from https://aka.ms/vs/17/release/vc_redist.x64.exe, and the checksums of https://aka.ms/vs/16/release/vc_redist.x64.exe and https://aka.ms/vs/17/release/vc_redist.x64.exe are different, so I assume winetricks still installs 2015-2019. 
see: https://github.com/clowd/Clowd.Squirrel/blob/dcbdbdb0cf3c825791cf384a6c1afa2eaa5e1a96/src/Squirrel/RuntimeInfo.cs#L490

Squirrel checks for VCRedist by getting the list of installed redistributables from the list of available uninstallers in the registry. 
see: https://github.com/clowd/Clowd.Squirrel/blob/dcbdbdb0cf3c825791cf384a6c1afa2eaa5e1a96/src/Squirrel/RuntimeInfo.cs#L442

As for the silent workaround to get around Squirrel hating Administrators, clearly that one doesn't work on the recent GE Proton/Wine builds. Since I was curious, I also tested the current release of XIVLauncher on our supported Lutris-7.1 wine build, and it installs XIVLauncher perfectly fine without the silent flag. [Phoronix claims](https://www.phoronix.com/scan.php?page=news_item&px=Wine-Staging-6.3) that wine-staging actually no longer reports itself as elevated by default any more, and I would assume that the way our ancient Squirrel was checking for elevation in recent GE Proton/Wine builds simply borked out.

reiichi001:
I'm tempted to add a vcrun2022 verb for winetricks, but I know I sure as hell don't want to maintain it.

Looking at the Squirrel checks though, I don't think it'd complain if it can't find anything specific to 2022 installed, as 2019 makes the same registry entries. 

If there's a test build, I can run through some common installs on Lutris using Fedora and Ubuntu, and then also do a fresh Proton install. ...the native steam users using Proton will probably be the most annoying, as many are on Proton-GE-6.21, which is based on Wine 6 of some form. ...so it might now have latest staging adjustments. I don't recall why they use that instead of Wine 7 codebases, but it's something we'll need to keep in mind and test for in the meantime.

reiichi001:
Ran a few tests on my current Linux distro of the moment.

winetricks dotnet48 vcrun2019 win10
-> prompted to install VC Runtime 2022
-> XIVLauncher would never run. I tried reinstalling a few times to no avail, which could be an issue for our current users as that would be a breaking change. Either we'd need to make some intermediate release on old SquirrelSetup that popped up for Linux users and told them to make a new prefix, or we may be best off targeting 2019 in general, but download the latest installer if we're going to handle that step for the user.

winetricks dotnet48 win10
-> prompted to install VCRuntime 2022
-> XIVLauncher did run
-> Not sure if this is a proper practice as the DLLs may or may not be registered correctly?

winetricks dotnet48 vcrun2022 win10
-> I did the thing and bolted in a vcrun2022 verb for winetricks
-> DLLs listed above are properly registered
-> No prompts as expected
-> XIVLauncher runs as expected

I didn't go so far as to patch up an install / launch with Dalamud as today was a bit hectic, but I can give those a run-through later to ensure everything still checks out. I'm not sure what the winetricks policy is on getting a PR merged, but since it's just a simple new verb, I'm optimistic. Otherwise, there's always the option of just not registering vcrun DLLs, but I'm a bit hesitant to sign off on that.

Dormanil:
Absolutely weird that XIVLauncher never ran/installed despite you being prompted to install VC2022 in the first case. 
It worked on my machine and VC2022 literally supplies the dlls named the same as VC2019, so I have no clue how that broke.

I am also hesitant on not registering vcrun dlls, because that means that we'd likely be defaulting to wine's built-in ones where those exist.

Dormanil:
Closing since no longer wanted/needed.

goaaats:
I still think this has merit for Windows, but I'll have to do a lot of testing

PR #881, opened 2022-04-13
Set GShade mode in registry when changed

Mortalitas:
Prevents Windows users from running into an issue where the GShade mode is reverted after GShade updates are installed.


goaaats:
Thanks, wasn't aware this was a thing

goaaats (approved):

PR #886, opened 2022-04-15
better command line args handling

ashkitten:
no more shelling out to `sh` except for custom command mode (which i haven't tested, but should function about the same as before)

### DIFF
--- a/src/XIVLauncher.Core/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Core/Compatibility/CompatibilityTools.cs
@@ -73,27 +73,28 @@ public class CompatibilityTools
 
     public void EnsurePrefix()
     {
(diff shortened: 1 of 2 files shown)


ashkitten:
so... testing "command" startup mode, it doesn't work. `WINEPREFIX` and `WINEDLLOVERRIDES` are specified *after* the `wine` command, so i'm not sure how this ever worked in the first place. that being the case, i don't think this broke anything that wasn't already broken.

i think for now it would be pragmatic to simply add `WINEPREFIX` and `WINEDLLOVERRIDES` to the `EnvironmentVariables` in both startup modes since that seems to be the sort of ergonomics it was going for in the first place? if the user specifies them in the command line itself, they'll be correctly overridden, but it may not be obvious that they're being set in the first place since they're not part of the `%COMMAND%` replacement.

goaaats (approved):

PR #892, opened 2022-04-17
Replace Assembly.Location and make single-file

Dormanil:
This commit replaces any instances of `Assembly.Location` in XIVLauncher.Core and XIVLauncher.Common with the appropriate alternative, usually `AppContext.BaseDirectory`. Any such uses of `Assembly.Location` would result in `string.Empty` when used in the context of a single-file executable. 

Yes, I am abusing a source generator to get the AssemblyVersion from the .csproj to somewhere accessible in the assembly, because `FileVersionInfo.GetVersionInfo()` just doesn't return any version info in a single-file executable, even when given a proper path to check.

And yes, I reenabled single-file and the runtime identifiers. Didn't break on my machine.


goaaats:
As we're technically also deploying on windows, we should probably still leave single file and RID out of the csproj. Can be specified in whatever CI step that makes the package though.

Dormanil:
Single-file only ever changes anything when using `dotnet publish`, and the specified RID needs to be supplied during build anyway, as specifying multiple supported RIDs in the .csproj doesn't specify anything when simply writing `dotnet build`. The only thing actually doing things is at all times is the `<SelfContained>true</SelfContained>`, which requires us to specify an RID during build/publish.

Dormanil:
In accordance with my prior comment, I have removed the `<SelfContained>true</SelfContained>` from the .csproj file. When publishing and under specification of an RID, self-contained is implied unless you specify the `--no-self-contained` compiler flag, so adding it in the project file just added unnecessary hassle.

Dormanil:
Closing this mess, gonna reopen with a cleaned up version.

PR #895, opened 2022-04-18
Implementation of basic Dalamud launcher for Linux/macOS

marzent:
This also includes a change to `IDalamudRunner`, although it should be fine on Windows on first glance.
I took the liberty to refactor all game and dalamud launch related stuff for Linux/macOS into `XIVLauncher.Common.Unix` similar to `XIVLauncher.Common.Windows`.
Also includes the latest steamworks dylib (not really part of Dalamud but got annoyed by the error message on every start :)


goaaats:
I'm surprised that it init'd steam with latest steamworks for you, it's actually based on v148a - on Windows, it crashes if you use the newest one

marzent:
Oupsies seems to be sth broken on Windows... will try to fix it up tomorrow 
Re:steam it also doesn't work when actually trying to login, so I should probably get the appropriate library (and actually test it this time)

marzent:
> Oupsies seems to be sth broken on Windows... will try to fix it up tomorrow

Seems to work (and inject) fine on Windows; build failure was caused by the recent `IndexedZiPatchInstaller` changes

> I'm surprised that it init'd steam with latest steamworks for you, it's actually based on v148a - on Windows, it crashes if you use the newest one

Thanks for the hint! Version 148a actually worked without issues.

src/XIVLauncher.Common.Unix/UnixGameRunner.cs
```diff
@@ -88,10 +100,18 @@ public LinuxGameRunner(LinuxStartupType startupType, string startupCommandLine,
...
+        Int32 gameProcessId = 0;
         helperProcess.BeginErrorReadLine();
-
-        string pid = helperProcess.StandardOutput.ReadToEnd();
-
-        return int.Parse(pid);
+        while (gameProcessId == 0)
+        {
+            Thread.Sleep(50);
+            gameProcessId = compatibility.GetProcessId("ffxiv_dx11.exe");
```
goaaats, line 108: How does this handle multiboxing? On Windows, we had to change this as it would always select the first process' PID.

src/XIVLauncher.Core/Components/MainPage/MainPage.cs
```diff
@@ -757,10 +760,22 @@ public async Task<Process> StartGameAndAddon(Launcher.LoginResult loginResult, b
...
-        // TODO(Linux): We need to translate the wine pid into the unix pid and go on here
-        if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+        if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+        {
+            var process = launched as Process;
+            await Task.Run(() => process!.WaitForExit()).ConfigureAwait(false);
+        }
+        else if (Environment.OSVersion.Platform == PlatformID.Unix)
+        {
+            for(; Program.CompatibilityTools.GetProcessId("ffxiv_dx11.exe") != 0;)
```
goaaats, line 771: Same thing here

src/XIVLauncher.Common.Unix/UnixGameRunner.cs
marzent: It would have to keep a list of pids where injection has been attempted I guess in order to properly deal with multiboxing, shouldn't be too hard to fix

goaaats: Does winedbg not let you go from unix pid to windows pid?

marzent: I tried, but couldn't find anything on it unfortunately

marzent: fixed with cf93cac2fcc7261cca3e5c3edd530b847d74118c

goaaats: So I guess we can merge this for now as it doesn't break Windows multiboxing, and whenever we get to build the pid resolver for wine, we can fix it in xlcore

PR #914, opened 2022-04-22
GameMode, DXVK_ASYNC, ESYNC and FSYNC support

Zi-SH:
Pull adds in GameMode, DXVK_ASYNC, and provisional ESYNC and FSYNC support. 

- New options added in the WINE tab.
- GameMode and FSYNC set only to show if the system supports them.
- ESYNC and FSYNC still have some oddities to work out, so their SettingEntry doesn't toggle them yet.
- Defaults should be consistent with current defaults. ESYNC is on by default, but specifically disabled in Runner.


src/XIVLauncher.Core/Components/SettingsPage/Tabs/SettingsTabWine.cs
```diff
@@ -23,6 +23,17 @@ public SettingsTabWine()
                 CheckVisibility = () => startupTypeSetting.Value == WineStartupType.Command
             },
 
+            new SettingsEntry<bool>("Enable GameMode", "Enable launching with GameMode optimizations.", () => Program.Config.GameModeEnabled ?? true, b => Program.Config.GameModeEnabled = b)
+            {
+                CheckVisibility = () => File.Exists("/usr/lib/libgamemodeauto.so.0")
```
goaaats, line 28: Do you think it'd be better to use CheckValidity here, to let people know that they have to install this first?

Zi-SH: Not going to lie, forgot it existed. I do think it'd be a preferable option to hiding it all together so I'll check it in the morning.

goaaats:
Did someone test this? Should we merge #915 first? @marzent

marzent:
> Did someone test this? Should we merge #915 first? @marzent

It makes sense to merge https://github.com/goatcorp/FFXIVQuickLauncher/pull/915  first, then ESYNC and FSYNC toggles should also be functional

goaaats:
It's a draft, just let me know when you get to finish it up and I'll have a go at it 👍

marzent:
Should be good now

Zi-SH:
Re-implemented all of the optimizations as they were previously, but utilizing the new CompatibilityTool. Unlocked ESync and FSync after testing multiple times.

goaaats (approved):

PR #915, opened 2022-04-22
Improve wine handling

marzent:
This makes it so that all Windows binaries are run through `CompatibilityTools.RunInPrefix`, in order to avoid running wine with different environment variables when called from different places...
Additionally the "command" option has been replaced with a "custom" one, since "command" will be broken in almost all cases; especially in combination with Dalamud. (In order to bring back most user customisation however it would be good to expose additional environment args somehow as in https://github.com/goatcorp/FFXIVQuickLauncher/pull/914)


src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
```diff
@@ -12,34 +13,40 @@ namespace XIVLauncher.Common.Unix.Compatibility;
...
-    public string WineServerPath => Path.Combine(toolDirectory.FullName, WINE_GE_RELEASE_NAME, "bin", "wineserver");
+    private const string WINE_TKG_RELEASE_URL = "https://github.com/Kron4ek/Wine-Builds/releases/download/7.6/wine-7.6-staging-tkg-amd64.tar.xz";
+    private const string WINE_TKG_RELEASE_NAME = "wine-7.6-staging-tkg-amd64";
 
-    public DirectoryInfo Prefix { get; private set; }
-    public DirectoryInfo DotnetRuntime { get; private set; }
     public bool IsToolReady { get; private set; }
 
-    public bool IsToolDownloaded => File.Exists(Wine64Path) && this.Prefix.Exists;
+    public readonly WineSettings wineSettings;
```
goaaats, line 24: This should be a property: `public WineSettings Settings { get; private set; }`
Looks good otherwise, thanks :)

goaaats (changes requested):

src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
marzent: Should be fixed now

goaaats (approved):

PR #920, opened 2022-04-24
Use HTTPS for xivapi.com

Playacem:
The xivapi docs [1] recommend using HTTPS to access the api.

 [1] https://xivapi.com/docs


goaaats:
This used to be an issue due to outdated crypto on Windows 7, fair to say that it should be fine nowadays

Issue #918, opened 2022-04-23
Core - FFXIV Config Folder Adjustments


rekyuu:
### Update disclaimer

- [X] Yes, I have checked and my request is not related to FFXIV updating and plugins not working correctly.

### What happened, what did not, what did you want to happen?

- Move FFXIV config folder to `XDG_CONFIG_HOME`
- Check to see if existing configs exist from previous installs (and ask if the user wants them to be copied/migrated to the new folder?)
- Implement a backup system for the config folder

### Version

Native/Core version (Linux)

### Platform

Linux (native)

Zi-SH:
Taking a look at this now, will adjust it as we discussed and enter a draft pull.

Figuera:
(First thank you, a native Linux launcher is something I have been looking for ages)

It should follow XDG Base Directory specification. See: https://wiki.archlinux.org/title/XDG_Base_Directory.

From what I gathered, this is where files should be located:

- $XDG_CONFIG_HOME (usually "~/.config/xlcore")
  - launcher.ini
  - dalamudConfig.json
  - ffxivConfig/* (if possible move screenshots to XDG_HOME_DATA but I assumed that is not possible)
  - pluginConfigs/*
 
 - $XDG_DATA_HOME (usually "~/.config/xlcore")
   - compatibilitytool/*
   - dalamud/*
   - dalamudAssets/*
   - devPlugins/*
   - installedPlugins/*
   - patch/*
   - runtime/*
   - wineprefix/*
   DalamudWineHelper.exe
   secrets.json
   
  - $XDG_CACHE_HOME (usually ~/.cache/xlcore")
    - All log files
   
Following the specification is nice, but I would be happy just by not having the directory on my home directory. So putting everything on `.config/` would be a fine solution for me.

Zi-SH:
As much as I'd love to make the application entirely XDG compliant (I hate unnecessary items in my home folder), there's a desire to keep the files together much like how XIVLauncher currently operates on Windows. XDG_DATA_HOME maybe a good final resting place for the entire .xlcore folder.

PR #924, opened 2022-04-25
Update to Configuration Storage

Zi-SH:
This is a direct modification to the changes by pull #908 , and as discussed in issue #918 .

- Introduce ConfigStorage, which brings defines configuration storage in line with the Roaming/XIVQuickLauncher folder on windows.
- Modification to the default FFXIV config folder location, with logic to attempt detection and usage of old configs to make moving less painful.
- Modification to the default dalamud.ini location

This pull is in response to my discussions with @rekyuu

### DIFF
--- a/src/XIVLauncher.Core/ConfigStorage.cs
+++ b/src/XIVLauncher.Core/ConfigStorage.cs
@@ -1,0 +1,81 @@
+namespace XIVLauncher.Core;
+
+public class ConfigStorage
+{
+    public DirectoryInfo XLConfigRoot { get; }
+    public DirectoryInfo GameConfigRoot { get; }
+
+    private readonly DirectoryInfo legacyXLConfigRoot;
+    private readonly DirectoryInfo legacyGameConfigRoot;
+    private readonly DirectoryInfo symlinkGameConfigs;
+    
+    public ConfigStorage(string appName, string configFolder)
+    {
+        // https://developers.redhat.com/blog/2018/11/07/dotnet-special-folder-api-linux
+        // Support for migrating old configs, to bring storage inline with Windows
+        this.legacyXLConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}"));
+        this.legacyGameConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}/ffxivConfig"));
+
+        this.XLConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "XIVLauncher"));
+        this.GameConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "XIVLauncher"));
+        this.symlinkGameConfigs = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "Documents/My Games/FINAL FANTASY XIV - A Realm Reborn"));
+
+        if (Directory.Exists(this.symlinkGameConfigs.FullName) && (!Directory.Exists(this.GameConfigRoot.FullName)))
+        {
+            // Copy directory, as moving it may break other mappings to the location
+            Directory.CreateDirectory(this.GameConfigRoot.FullName);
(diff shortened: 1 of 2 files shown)


src/XIVLauncher.Core/ConfigStorage.cs
```diff
@@ -0,0 +1,81 @@
...
+    private readonly DirectoryInfo legacyXLConfigRoot;
+    private readonly DirectoryInfo legacyGameConfigRoot;
+    private readonly DirectoryInfo symlinkGameConfigs;
+    
+    public ConfigStorage(string appName, string configFolder)
+    {
+        // https://developers.redhat.com/blog/2018/11/07/dotnet-special-folder-api-linux
+        // Support for migrating old configs, to bring storage inline with Windows
+        this.legacyXLConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}"));
+        this.legacyGameConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}/ffxivConfig"));
```
goaaats, line 17: I don't think we actually need this, seeing as xlcore isn't released yet and this stuff has been on master for less than a week... Better to make this less complex.

src/XIVLauncher.Core/ConfigStorage.cs
```diff
@@ -0,0 +1,81 @@
...
+            File.Copy(this.symlinkGameConfigs.FullName + "/FFXIV.cfg", this.GameConfigRoot + "/FFXIV.cfg");
+            File.Copy(this.symlinkGameConfigs.FullName + "/FFXIV_BOOT.cfg", this.GameConfigRoot + "/FFXIV_BOOT.cfg");
+        }
+        else
+        {
+            if (Directory.Exists(legacyGameConfigRoot.FullName) && (!Directory.Exists(this.GameConfigRoot.FullName)))
+                Directory.Move(legacyXLConfigRoot.FullName, XLConfigRoot.FullName);
+        }
+
+        if (!this.GameConfigRoot.Exists)
```
goaaats, line 36: Should this not be before L23? What happens if someone has configs in documents, but never used xlcore before?

src/XIVLauncher.Core/ConfigStorage.cs
```diff
@@ -0,0 +1,81 @@
...
+    
+    public ConfigStorage(string appName, string configFolder)
+    {
+        // https://developers.redhat.com/blog/2018/11/07/dotnet-special-folder-api-linux
+        // Support for migrating old configs, to bring storage inline with Windows
+        this.legacyXLConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}"));
+        this.legacyGameConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}/ffxivConfig"));
+
+        this.XLConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "XIVLauncher"));
+        this.GameConfigRoot = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "XIVLauncher"));
```
goaaats, line 20: Why are these not the same? What's the use case, especially seeing as "ffxivConfig" is appended to the end below? Aren't the copies going to the wrong place?

goaaats, line 20: Why are these not the same? What's the use case, especially seeing as "ffxivConfig" is appended to the end below? Aren't the copies going to the wrong place?

goaaats, line 20: Why are these not the same? What's the use case, especially seeing as "ffxivConfig" is appended to the end below? Aren't the copies going to the wrong place?

goaaats (changes requested):

src/XIVLauncher.Core/ConfigStorage.cs
Zi-SH: The check for `Config.GameConfigPath` in `Program.cs` is specifying `ffxivConfig`. Adding `ffxivConfig` in for GameConfigRoot had it go to `~/.config/XIVQuickLauncher/ffxivConfig/ffxivConfig`.

I adjusted `GameCoreRoot` to include `ffxivConfig`, and adjusted `GetGameConfigFolder` to return that folder if called without parameters.

Issue #921, opened 2022-04-24
[Core] - "-all" set as default WINEDEBUG (Low Priority)


Zi-SH:
### Update disclaimer

- [X] Yes, I have checked and my issue is not related to FFXIV updating and plugins not working correctly.

### What did you do?

Low priority bug for Settings UI - 
Core is showing `-all` in the WINEDEBUG field in the settings. 

Rather than set the default of the field to -all, it maybe preferable to check if `Config.WineDebugVars` is empty and if so apply `-all` at launch.

### Platform

Linux (native)

### Wine/Proton runner version

N/A

PR #925, opened 2022-04-25
Adjusted DebugVar default, set -all if emtpy 

Zi-SH:
Rather than set `-all` as the default `WINEDEBUG`, which shows in the Settings, sets it in WineSettings if debugVars is empty.

This pull is in reference to #921

### DIFF
--- a/src/XIVLauncher.Common.Unix/Compatibility/WineSettings.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/WineSettings.cs
@@ -26,7 +26,7 @@ public class WineSettings
     {
         this.StartupType = startupType ?? WineStartupType.Custom;
         this.CustomBinPath = customBinPath;
-        this.DebugVars = debugVars;
+        this.DebugVars = string.IsNullOrEmpty(debugVars) ? "-all" : debugVars;
         this.LogFile = logFile;
         this.Prefix = prefix;
     }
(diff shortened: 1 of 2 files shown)


marzent:
This change is equivalent behaviour to what it does currently though and does not address #921...

Zi-SH:
Re: #921 
> Low priority bug for Settings UI -
> Core is showing -all in the WINEDEBUG field in the settings.

> Rather than set the default of the field to -all, it maybe preferable to check if Config.WineDebugVars is empty and if so apply -> all at launch.

The change:

- Hides "-all" from the UI by removing it from the config as the default WINEDEBUG setting on launch
- Sets DebugVars to "-all" if DebugVars is empty, which is just Config.WineDebugVars passed as a parameter

Is there something I'm missing?

marzent:
Ah seems I misread #921 as the WINEDEBUG field not saving but rather reverting to "-all" on every launch.

To be honest I think it would be better to have a 1:1 correspondence between the field value and the environment variable, especially since an empty field prints ERR and FIXME messages which is expected behaviour from wine...

> Rather than set the default of the field to -all, it maybe preferable to check if Config.WineDebugVars is empty and if so apply - all at launch.

Currently the WINEDBUG variable will not be applied at launch from the config but is hardcoded to "" in the PR, which then gets later replaced by "-all" every time, resulting in effectively the same behaviour as before, without actually giving the user any indication they are currently running with this debug channel.

Zi-SH:
You raise a good point about it being the default **expected** behavior of WINE. My intention was to have WINE default to a cleaner logging output using the `-all` and not have it constantly in the WINEDEBUG lest users think something was going on, but there would need to be at least some indication that the currently logging level is not the default.

I'll close this PR, and if I can come up with a better more transparent open I'll open a new PR.

PR #934, opened 2022-05-02
Move unnecessary files into separate directory on repair

Soreepeong:
All files that did not come with the game installation will be moved into a subdirectory called "repair_recycler" on repair.

image


goaaats (commented):
chore: add note about GShade reinstalling

PR #949, opened 2022-05-12
refactor: abstract out Launcher > ILauncher/SqexLauncher

avafloww:
... and adds barebones ActozLauncher / ShandaLauncher for future use (or not).

This refactor helps to prepare XL for the possibility of supporting logins and patching for KR/CN FFXIV, although there would be a lot more work/complexity there (especially around Dalamud and the different game versions). In reality, it's mainly to be able to track KR/CN game versions in Thaliak, though of course I'd love to try to clean things up and upstream these changes.

### DIFF
--- a/src/XIVLauncher.Common/Game/Exceptions/VersionCheckLoginException.cs
+++ b/src/XIVLauncher.Common/Game/Exceptions/VersionCheckLoginException.cs
@@ -1,0 +1,14 @@
+using System;
+
+namespace XIVLauncher.Common.Game.Exceptions;
+
+public class VersionCheckLoginException : Exception
+{
+    public LoginState State { get; }
+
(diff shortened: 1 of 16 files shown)


marzent:
Would it be more reasonable to abstract out the Steam Launcher into its own class?

Especially considering there are three different platforms for the international version of FFXIV and a bool is a bit awkward to deal with that… (Something like a `SquexLauncher` base class maybe?)

avafloww:
@marzent Yeah, I think that makes sense for dealing with the weirdness of Steam, good call! I'll do that.

Bluefissure:
Hi, I think this pull will be a great help for what we're doing in [XIVLauncherCN](https://github.com/ottercorp/FFXIVQuickLauncher), and we'll be grateful if it can be merged to our upstream soon!

Issue #963, opened 2022-05-16
Access denied on new injection method


Aireil:
### Update disclaimer

- [X] Yes, I have checked and my issue is not related to FFXIV updating and plugins not working correctly.

### What did you do?

On normal launcher Windows, when using new injection method, access denied on `gameProcess.ProcessName` [here.](https://github.com/goatcorp/FFXIVQuickLauncher/blob/master/src/XIVLauncher.Common.Windows/WindowsDalamudRunner.cs#L32)

Regression from 55c25c3a59606d84449854525a24bc5d5a19784d.

Good luck @marzent ♥

### Platform

Windows

Issue #957, opened 2022-05-14
[xlcore]  Alternative Game Config Path parsed incorrectly in AUR build(s)


taylor85345:
### Update disclaimer

- [X] Yes, I have checked and my issue is not related to FFXIV updating and plugins not working correctly.

### What did you do?

1. Install xivlauncher-git from AUR.
2. Launch and set Game Path and Game Config Path to directories outside of `/home/<user>/.xlcore`, containing existing FFXIV game files and config files respectively
3. Game launches successfully from existing game files, but does not load config files. New config directory is generated at `<Game Path>/game/ 
4.  UserPath=Z:/<Game Config Path>`

For example, I have the following directories in my config:

Game Path: `/home/taylor/ffxiv/SquareEnix/FINAL FANTASY XIV - A Realm Reborn/`
Game Config Path: `/home/taylor/ffxiv/xivconf/`

Ultimately, the launcher creates a new directory with new configs in the following location:
`/home/taylor/ffxiv/SquareEnix/FINAL FANTASY XIV - A Realm Reborn/game/  UserPath=Z:/home/taylor/ffxiv/xivconf`

Side Note, the Game Config Path field also seems to be unable to handle spaces in the path entry. My configs were originally at `/home/taylor/ffxiv/FINAL FANTASY XIV - A Realm Reborn`, and the config path that it generated was `/home/taylor/ffxiv/SquareEnix/FINAL FANTASY XIV - A Realm Reborn/game/  UserPath=Z:/home/taylor/ffxiv/Reborn`

### Platform

Linux (native)

### Wine/Proton runner version

Default (Managed bu XIVLauncher)

PR #965, opened 2022-05-18
Use Dalamuds new console interface to launch the game

marzent:
This makes use of https://github.com/goatcorp/Dalamud/pull/811 in order to inject Dalamud ~~so that the ACL fix component can be removed from XL~~.

~~To compensate for the use case of an ACL fixed launch without Dalamud, a new `DalamudLoadMethod` is introduced that is currently exposed in xl-core and could, if need be, added to WPF XL as well.~~

This should fix fake login on non-Windows systems on xl-core and https://github.com/goatcorp/FFXIVQuickLauncher/issues/957.

In Addition since only `Process` objects and no PIDs are passed around anymore this should also fix https://github.com/goatcorp/FFXIVQuickLauncher/issues/963, @Aireil would be much appreciated if you could give this a spin ♥


Aireil:
It started for me, fixing #963, didn't test anything else 👍 .

goaaats:
XL not acl-fixing when Dalamud is disabled isn't really an option for me, because that's how I dev. I honestly believe that it's best to leave it to XL.

marzent:
Just thought it would help with AV targeting XL maybe, could be left in as-is if the launch through Dalamud but no injection mode is not good enough (well and there is still no GUI for that in the WPF version)

goaaats:
We'd have to change our distribution model for dalamud then, because at the moment, it's entirely ignored on patches. It's good to keep this in mind but we should try to fix the original issue first, imo.

marzent:
Ah I see, haven’t thought of that…
Would using the ACL fix on the Windows side of things be enough or should DalamudWineHelper also be put back to launch with fixed ACLs without dalamud

goaaats (approved):

PR #967, opened 2022-05-21
Fix UserPath with encrypted args

marzent:
Only escape with double quotes when not encrypting args


goaaats:
Could you describe what the problem was?

marzent:
I believe SE has its own way of encoding spaces when encrypting arguments by doubling them and that somehow interferes with quoting, which is necessary though when passing in a path containing spaces unencrypted

goaaats:
That makes sense, I think we've run into this before

PR #990, opened 2022-05-26
Implementing repair for xlcore

marzent:
This adds a new class `IndexedZiPatchIndexLocalInstaller` and fixes an issue where the patch index files were not being read correctly on non-windows platforms.


marzent:
Should work around a breaking change in .NET 6 and up concerning the full reading of a `DeflateStream` without using a third party library now.

goaaats (approved):

PR #1011, opened 2022-05-29
Implement FPS Limiter (DXVK_FRAME_RATE)

Zi-SH:
Resubmitting #948 with fix for merge conflict and a few changes for Linux-only visibility.

I've been playing with this setting for a while and I've not had any issues with DXVK's frame limiter.

### DIFF
--- a/src/XIVLauncher.Common.Unix/Compatibility/WineSettings.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/WineSettings.cs
@@ -19,15 +19,18 @@ public class WineSettings
     public string EsyncOn { get; private set; }
     public string FsyncOn { get; private set; }
 
+    public string DxvkFPSLimit { get; private set; }
+
     public string DebugVars { get; private set; }
     public FileInfo LogFile { get; private set; }
 
     public DirectoryInfo Prefix { get; private set; }
 
-    public WineSettings(WineStartupType? startupType, string customBinPath, string debugVars, FileInfo logFile, DirectoryInfo prefix, bool? esyncOn, bool? fsyncOn)
+    public WineSettings(WineStartupType? startupType, string customBinPath, string debugVars, FileInfo logFile, DirectoryInfo prefix, bool? esyncOn, bool? fsyncOn, string dxvkFPSLimit)
     {
         this.StartupType = startupType ?? WineStartupType.Custom;
         this.CustomBinPath = customBinPath;
+        this.DxvkFPSLimit = dxvkFPSLimit;
         this.EsyncOn = (esyncOn ?? false) ? "1" : "0";
         this.FsyncOn = (fsyncOn ?? false) ? "1" : "0";
         this.DebugVars = debugVars;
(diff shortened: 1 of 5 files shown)


src/XIVLauncher.Core/Components/SettingsPage/Tabs/SettingsTabWine.cs
```diff
@@ -54,7 +54,6 @@ public SettingsTabWine()
             },
             new SettingsEntry<string>("DXVK Frame Limit", "Configure how many frames DXVK should be limited to. Set to 0 for unlimited.", () => Program.Config.DxvkFrameRate ?? "0", s => Program.Config.DxvkFrameRate = s)
             {
-                CheckVisibility = () => RuntimeInformation.IsOSPlatform(OSPlatform.Linux),
```
Zi-SH: After discussion with marzent about DXVK on Windows, I've opted to remove Linux-only visibility/

src/XIVLauncher.Common.Unix/Compatibility/WineSettings.cs
```diff
@@ -19,15 +19,18 @@ public class WineSettings
     public string EsyncOn { get; private set; }
     public string FsyncOn { get; private set; }
 
+    public string DxvkFPSLimit { get; private set; }
```
marzent, line 22: Wouldn't it make more sense to store an fps limit as an integer

src/XIVLauncher.Common.Unix/Compatibility/WineSettings.cs
```diff
@@ -19,15 +19,18 @@ public class WineSettings
...
+
     public string DebugVars { get; private set; }
     public FileInfo LogFile { get; private set; }
 
     public DirectoryInfo Prefix { get; private set; }
 
-    public WineSettings(WineStartupType? startupType, string customBinPath, string debugVars, FileInfo logFile, DirectoryInfo prefix, bool? esyncOn, bool? fsyncOn)
+    public WineSettings(WineStartupType? startupType, string customBinPath, string debugVars, FileInfo logFile, DirectoryInfo prefix, bool? esyncOn, bool? fsyncOn, string dxvkFPSLimit)
```
marzent, line 29: This is not a wine setting and should probably go into the appropriately named `Dxvk` class

src/XIVLauncher.Core/Components/SettingsPage/Tabs/SettingsTabWine.cs
```diff
@@ -51,7 +52,10 @@ public SettingsTabWine()
                     return null;
                 }
             },
-
+            new SettingsEntry<string>("DXVK Frame Limit", "Configure how many frames DXVK should be limited to. Set to 0 for unlimited.", () => Program.Config.DxvkFrameRate ?? "0", s => Program.Config.DxvkFrameRate = s)
+            {
+                CheckValidity = s => Regex.IsMatch(s, @"^\d{1,3}$") ? null : "Please specify a valid integer below 999."
```
marzent, line 57: Regex is a bit overkill to check for a non-negative integer; afaik dxvk can also limit to more than 999fps

src/XIVLauncher.Core/Configuration/ILauncherConfig.cs
```diff
@@ -72,6 +72,8 @@ public interface ILauncherConfig
 
     public bool? FSyncEnabled { get; set; }
 
+    public string DxvkFrameRate { get; set; }
```
marzent, line 75: Same as earlier this does not need to be a string

marzent (changes requested):

Zi-SH:
I don't have access to my system for another week, but I'll review it after I return.

PR #1014, opened 2022-05-31
Fixes #1013

### LINKED ISSUE #1013: Plugin disable/enable in launcher settings should be localized
### Update disclaimer

- [X] Yes, I have checked and my issue is not related to FFXIV updating and plugins not working correctly.

### What did you do?

Disable plugins in XIVLauncher settings and enable, when using XIVLauncher languages other than English, the enable doesn't work.

Possible cause: `(disabled)` should be localized at https://github.com/goatcorp/FFXIVQuickLauncher/blob/master/src/XIVLauncher/Windows/SettingsControl.xaml.cs#L386

### Platform

Windows

### Wine/Proton runner version

_No response_

### Relevant log output

_No response_

Bluefissure:
Fixes https://github.com/goatcorp/FFXIVQuickLauncher/issues/1013

### DIFF
--- a/src/XIVLauncher/Windows/SettingsControl.xaml.cs
+++ b/src/XIVLauncher/Windows/SettingsControl.xaml.cs
@@ -383,7 +383,7 @@ namespace XIVLauncher.Windows
                 return;
             }
 
-            if (PluginListView.SelectedValue.ToString().Contains("(disabled)")) //If it's disabled...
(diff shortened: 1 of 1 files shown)


goaaats:
Thanks, that view is a mess and definitely needs to be redone... It doesn't correctly reflect plugin state anymore as they now store this info in the manifest.

Issue #909, opened 2022-04-21
[xlcore] add a toggle for Esync, Fsync, feral gamemode and mangohud


Caliel666:
### Update disclaimer

- [X] Yes, I have checked and my request is not related to FFXIV updating and plugins not working correctly.

### What happened, what did not, what did you want to happen?

the env are `WINEESYNC=1` `WINEFSYNC=1` `MANGOHUD=1` to enable them, note that to enable fsync one must have a patched kernel

as for gamemode it's a wrapper program, using gamemoderun after the env's and before the game launching

probably also wanna look into adding the env `WINEFSYNC_FUTEX2=1`

### Version

Native/Core version (Linux)

### Platform

Linux (native)

varris1:
A toggle for futex2 shouldn't be needed, as its automatically used with wine-tkg while using kernel 5.17+

Zi-SH:
This should be resolved as of #914 . @varris1 is correct in that futex2 will be enabled on newer kernels automatically. No need to provide options for both. 

As for MangoHUD, DXVK has a built-in frame limiter and there is already a DXVK overlay toggle. I believe implementing MangoHUD into XLCore would be redundant.

PR #1020, opened 2022-06-04
Added MangoHud support

Zi-SH:
Added MangoHud as requested by #909 and another user in the Discord.

This feature is excluded from Flatpak like Feral's Gamemode, due to complications with the sandboxing and XDG portals.


dlm21:
For anyone waiting on this, you can also just call it from the CLI `mangohud XIVLauncher.Core`

PR #1055, opened 2022-07-03
Add mostly-working integrity check to xlcore

reiichi001:
This add an Integrity Check button to XLCore's settings, right below the gamepath text field.

Under the hood, it runs the normal integrity process, but then swaps the path delimiters to backslashes, so they'll line up with a Windows-made integrity report.

I'm not able to make a modal display during the check, or note which file is being checked in a modal and need some help there. If there's a better way to do it, lemme know or please help me fix my code!

After the check has ran, a modal will be displayed with some info on the results, directly copied from the XAML code, so Loc can probably be matched up later to reduce work.

### DIFF
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Diagnostics;
@@ -57,7 +57,14 @@ public class CompatibilityTools
         this.toolDirectory = new DirectoryInfo(Path.Combine(toolsFolder.FullName, "beta"));
         this.dxvkDirectory = new DirectoryInfo(Path.Combine(toolsFolder.FullName, "dxvk"));
 
-        this.logWriter = new StreamWriter(wineSettings.LogFile.FullName);
+        try
+        {
+            this.logWriter = new StreamWriter(wineSettings.LogFile.FullName);
+        }
+        catch (IOException e)
+        {
+            // For some reason, this is called on Windows, and then it crashes xlcore because wine.log is considered open by another process already.
+        }
 
         if (wineSettings.StartupType == WineStartupType.Managed)
         {
--- a/src/XIVLauncher.Common/Game/IntegrityCheck.cs
+++ b/src/XIVLauncher.Common/Game/IntegrityCheck.cs
@@ -7,6 +7,7 @@ using System.Security.Cryptography;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Serilog;
+using XIVLauncher.Common.Util;
 
 namespace XIVLauncher.Common.Game
 {
@@ -110,7 +111,7 @@ namespace XIVLauncher.Common.Game
                 var relativePath = file.FullName.Substring(rootDirectory.Length);
 
                 // for unix compatibility with windows-generated integrity
-                if (!OperatingSystem.IsWindows())
+                if (PlatformHelpers.GetPlatform() == Platform.Win32 || PlatformHelpers.GetPlatform() ==  Platform.Win32OnLinux)
                 {
 #if DEBUG
                     Log.Debug($"{relativePath} swapping to {relativePath.Replace("/", "\\")}");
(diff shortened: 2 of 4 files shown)


src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
```diff
@@ -57,7 +57,14 @@ public CompatibilityTools(WineSettings wineSettings, Dxvk.DxvkHudType hudType, b
...
         this.dxvkDirectory = new DirectoryInfo(Path.Combine(toolsFolder.FullName, "dxvk"));
 
-        this.logWriter = new StreamWriter(wineSettings.LogFile.FullName);
+        try
+        {
+            this.logWriter = new StreamWriter(wineSettings.LogFile.FullName);
+        }
+        catch (IOException e)
+        {
+            // For some reason, this is called on Windows, and then it crashes xlcore because wine.log is considered open by another process already.
```
marzent, line 66: Yeah I also noticed that when testing repair on windows and made sure it doesn’t happen here https://github.com/goatcorp/FFXIVQuickLauncher/blob/0e7473a271ddb41a7b763fafaf5c64a0a033bf18/src/XIVLauncher.Core/Components/SettingsPage/Tabs/SettingsTabWine.cs#L111

Probably even cleaner would be to not save settings pages that are hidden

src/XIVLauncher.Common/Game/IntegrityCheck.cs
```diff
@@ -108,6 +110,16 @@ private static IntegrityCheckResult DownloadIntegrityCheckForVersion(string game
             {
                 var relativePath = file.FullName.Substring(rootDirectory.Length);
 
+                // for unix compatibility with windows-generated integrity
+                if (PlatformHelpers.GetPlatform() == Platform.Win32 || PlatformHelpers.GetPlatform() ==  Platform.Win32OnLinux)
```
marzent, line 114: Shouldn’t this be the opposite?

marzent, line 114: Shouldn’t this be the opposite?

src/XIVLauncher.Common/Game/IntegrityCheck.cs
```diff
@@ -108,6 +110,16 @@ private static IntegrityCheckResult DownloadIntegrityCheckForVersion(string game
...
+                {
+#if DEBUG
+                    Log.Debug($"{relativePath} swapping to {relativePath.Replace("/", "\\")}");
+#endif
+                    relativePath = relativePath.Replace("/", "\\");
```
marzent, line 119: If all path replacements were from `Path.DirectorySeparatorChar` to `"\\"` they could always be made to avoid issues with the platform detection (assuming all paths are supposed to be canonically DOS-style, everything could also be made to use `/` )

src/XIVLauncher.Common/Game/IntegrityCheck.cs
reiichi001: Probably yes! Although...it ran. I'm gonna add some new debug checks and adjust accordingly. Will update the PR with a fix.

src/XIVLauncher.Common/Game/IntegrityCheck.cs
reiichi001: This is also definitely an option (as opposed to doing a replacement on the string). ...given that the bulk of users are on Windows and past integrity files are built that way, we'd probably be best sticking to backslashes to maintain compatibility (for the rare cases where someone may want to verify an old client version).

src/XIVLauncher.Common/Game/IntegrityCheck.cs
reiichi001: Update: so when I copy my build from Windows and run it with .Net, GetPlatform() resolves to Win32, even on Linux. I'm assuming this is a compile-time thing, but that means it probably shouldn't be trusted as the source of truth. (So the code above was technically working fine for my testing, because it was wrong from the beginning)

reiichi001: Thinking about it, it honestly doesn't need a platform check. We can just make the replacement always, as Windows doesn't really care about separators, but there isn't a good way to maintain compatibility with the old integrities while moving everything to `/`.  Kinda janky, but certainly not the worst possible way to handle this.

PR #1066, opened 2022-07-10
Bump version of Mac App Zip

marzent:
Seems SE has taken off the old 1.0.7 link from their origin server and it is starting to 404 for some regions (1.0.8 download speeds will probably be slow in the beginning but should stabilize after a while)


goaaats:
Thanks for looking out for this!

PR #1079, opened 2022-07-25
Add ability to reset Dalamud and plugins from XIVlauncher before running the game

urdenster:
Added option to hold Alt+Ctrl while booting Launcher to delete dalamud and plugins related folders.


Aireil:
This is quite the operation to do without any confirmation knowing holding ctrl or shift is already used by the launcher (someone could confuse alt and shift quite easily).
There were talks of better troubleshooting options, maybe it would fit there, but this is not a good idea to link this to hotkeys in my opinion.

NotNite:
It should, at the very least, prompt the user.

PR #1098, opened 2022-08-11
Update the SettingsTabWine to add a Flatpak check for gamemode.

orowith2os:
This allows GameMode to work with XIVLauncher inside of Flatpak. I'll rework the PR on the Flatpak manifest to account for this change, when a new version comes out that has this change.

### DIFF
--- a/src/XIVLauncher.Core/Components/SettingsPage/Tabs/SettingsTabWine.cs
+++ b/src/XIVLauncher.Core/Components/SettingsPage/Tabs/SettingsTabWine.cs
@@ -29,9 +29,7 @@ public class SettingsTabWine : SettingsTab
                 CheckValidity = b =>
                 {
 
-                     if (b == true && !FLATPAK && !File.Exists("/usr/lib/libgamemodeauto.so.0"))
-                         return "GameMode not detected.";
-                     else if (b == true && FLATPAK && !File.Exists("/app/lib/libgamemodeauto.so.0"))
+                     if (b == true && (!File.Exists("/usr/lib/libgamemodeauto.so.0") && !File.Exists("/app/lib/libgamemodeauto.so.0")))
                          return "GameMode not detected.";
 
                     return null;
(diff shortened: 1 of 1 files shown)


NotNite:
Why repeat a block? Feels like it makes more sense to do something like
```cs
#if FLATPAK
var gamemodePath = "/app/lib/libgamemodeauto.so.0";
#elseif
var gamemodePath = "/usr/lib/libgamemodeauto.so.0";
#endif
```
IMO.

orowith2os:
I'm not all that familiar with C#, so I went the easiest route. I'll apply that way of checking, and see if it works.

orowith2os:
Maybe something like this would be the best way of going about the check?

```            new SettingsEntry<bool>("Enable Feral's GameMode", "Enable launching with Feral Interactive's GameMode CPU optimizations.", () => Program.Config.GameModeEnabled ?? true, b => Program.Config.GameModeEnabled = b)
            {
                CheckVisibility = () => RuntimeInformation.IsOSPlatform(OSPlatform.Linux),
                CheckValidity = b =>
                {

                     if (b == true && (!File.Exists("/usr/lib/libgamemodeauto.so.0") || !File.Exists("/app/lib/libgamemodeauto.so.0")))
                         return "GameMode not detected.";

                    return null;
                }
            },
```

orowith2os:
hm, it seems the way I would do it has some issues, where if one of them doesn't exist, it doesn't work at all.

orowith2os:
Alright, I think I figured something out:
```
                     if (b == true && !FLATPAK && !File.Exists("/usr/lib/libgamemodeauto.so.0"))
                         return "GameMode not detected.";
                     else if (b == true && FLATPAK && !File.Exists("/app/lib/libgamemodeauto.so.0"))
                         return "GameMode not detected.";

                    return null;```

orowith2os:
Alright, I made a change, removes any Flatpak checks, and instead checks both /app/lib and /usr/lib for the appropriate gamemode files. It seems to build without any issues, and the logic for the check seems sound: If on Linux, and no gamemode files exist, it returns null.

orowith2os:
And verified to check that it works:
image
I'd say it's ready to merge now.

PR #1100, opened 2022-08-12
Removes the Flatpak limit for gamemode, small lib edit

orowith2os:
It now checks /app/lib/ and /usr/lib/, since the Flatpak needs to have user-added deps in /app/.

I cancelled the previous PR, as there were unnecessary commits, and this would clean up the history a bit.


orowith2os:
This is all that would need to be changed, it does not affect LD_PRELOAD whatsoever, at least from my testing.

orowith2os:
Got a suggestion from a user on Discord; Maybe check for the `gamemoderun` command from the `$PATH` rather than checking specific paths. This would make it less of a hack in the future to support things like MangoHUD.

marzent (approved):
LGTM

goaaats (approved):

PR #1160, opened 2022-09-24
Add generate tspack option to xlcore

reiichi001:


rankynbass:
I forked the master and merged this in. I could compile the result, but it wouldn't log in, instead giving that error about not being able to check the version. I then tried it without the "Define Linux as a platform" commit, and that compiled *and* logged in fine. I was also able to use the "Generate Troubleshooting Pack" button on the about page.

rankynbass:
That fixes it. Builds, installs, logs in, and can get in game. Tspack can be created from the About tab. Attached one as proof, although I had to add .zip to the end to get it to attach.
[trouble-20221001021635.tspack.zip](https://github.com/goatcorp/FFXIVQuickLauncher/files/9691217/trouble-20221001021635.tspack.zip)

goaaats (approved):

PR #1167, opened 2022-10-04
Allow use of WineD3D with environment variable. (Linux)

rankynbass:
This is a simple patch to allow the use of WineD3D instead of Vulkan. The USE_WINED3D environment variable can be passed to explicitly enable or disable WINED3D. The latter is useless at the moment, but perhaps in the future WineD3D will be included as a setting in the UI, so I put in code to deal with it. Valid values are 0 or false to disable, 1 or true to enable.

Usage example in Steam: `USE_WINED3D=1 %command% --parent-expose-pids --parent-share-pids --parent-pid=1`

I don't know how to build the flatpak, so I haven't been able to test that, but it works fine with a native build. Performance isn't terribly good on an AMD 6700XT, but it might work better for nVidia or Intel. It might also provide a solution if someone is trying to use an older graphics card that doesn't support Vulkan.

Here's a screenshot using mangohud to prove it's running:
WineD3D_working

### DIFF
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -156,6 +156,17 @@ public class CompatibilityTools
         psi.UseShellExecute = false;
         psi.WorkingDirectory = workingDirectory;
 
+        // Environment variable USE_WINED3D can be used to set wineD3D
+        string useWineD3D = Environment.GetEnvironmentVariable("USE_WINED3D") ?? "";
+        if (useWineD3D.Equals("1") || useWineD3D.ToLower().Equals("true"))
+        {
+            wineD3D = true;
+        }
+        else if (useWineD3D.Equals("0") || useWineD3D.ToLower().Equals("false"))
+        {
+            wineD3D = false;
+        }
+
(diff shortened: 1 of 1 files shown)


goaaats:
Is `USE_WINED3D` a "standardized" variable for this? Usually all xivlauncher variables are prefixed with `XL_`, you can find them in EnvironmentSettings.cs.

marzent:
Is it maybe possible to add this like a as a checkbox somewhere close to the dxvk settings?
In it’s current implementation it would override the wineD3D arg passed which is a bit unintuitive when in the future we would (hopefully) have auto launch functionality where you can select entries to be started with either D3D backend independently of each other…

Alternatively incorporating @goaaats suggestion something like `XL_FORCE_WINED3D` to make it clear that something is (potentially) being ignored here.

rankynbass:
#goats: no, USE_WINED3D is not a standard, although proton uses something similar. Could easily change it to XL_FORCE_WINED3D like marzent suggested. Also, I now see how it's done now with the EnvironmentSettings.cs. I could add a line in there, and then check that variable instead. I'm not sure how to edit this PR with those changes, though. Should I just close this and do a new one with the changes?

rankynbass:
Closing this. I've written a better patch that also prevents some crashes if you accidentally put invalid values in for env variables.

PR #1189, opened 2022-10-29
(Linux) Add a XL_FORCE_WINED3D variable, and fix a crash with bad env variable values

rankynbass:
Two changes:
1. Add an environment variable, XL_FORCE_WINED3D, to force the use of wined3d instead of dxvk. Name changed due to suggestion by marzent 
2. Change the function for CheckEnvBool(). Currently, it will crash the launcher if you set one of the XL environment variables to something other than "true" or "false" or nothing. So XL_WINEONLINUX=1 will crash. This patch will now make CheckEnvBool() evaluate to true with values of true, 1, on, and yes. Everything else will evaluate to false, allowing the launcher to ignore bad values instead of crashing.


rankynbass:
Closing this commit for now. It's not that important, and if PR#10 on XL.Core gets approved it can be done there instead.

PR #1205, opened 2022-12-10
[xlcore][feature] Dxvk settings rework

rankynbass:
Edit: Links updated for building with marzent's suggestions added.

This PR will add the necessary framework to XIVLauncher.Common.Unix to allow XL Core to add several new features around DXVK. This is only the first part of the patch, but it needs to be in place before XL Core can be updated.

* Moves all DXVK-related settings to a new class - DxvkSettings.
* Adds support for DXVK version switching. Currently this includes 1.10.1 - 1.10.3, and 2.0. Default is 1.10.1, and it pulls from the dxvk-async repo just like before.
* Adds support for DXVK_FRAMERATE_LIMIT. Setting the value to 0 sets framerate to unlimited.
* Adds support for custom DXVK Hud strings. This includes verifying that you only used supported options.
* Adds support for MangoHud
* Actually does some checking on the file paths for MangoHud, so that you acutally have to pass it a file for MangoHud Custom.
* Includes sensible fallbacks in the case that these changes are built against the current XL Core. Only the new DXVK hud options and MangoHud will show up, but it will build and run.
* Change the 4th argument in the constructor method in Launcher.cs to have a default value of null; this is just to allow XL Core to build without changes. 

I've been working on this for a while, trying to get it into a place where it would work in a flatpak environment without hanging or crashing. It's finally working now.

You can try it out for yourself by following the instructions here: [XL Core Testing Flatpak](https://github.com/rankynbass/dev.goats.testing.xivlauncher/tree/new-dxvk-settings), or build it by cloning [my fork of XL Core](https://github.com/rankynbass/XIVLauncher.Core/tree/xlcore-fixes-plus-dxvk) and checking out the xlcore-fixes-plus-dxvk branch, and doing a `dotnet publish`

The `master` branch includes all the XL Core updates I'm working on, and the 1.0.2+new-dxvk-settings branch, is there to prove that it can be built against XL Core with no changes other than updating the submodule.

Here's some screenshots of how it looks (with XL Core patches).

Wine Tab:
dxvk-rework-wine-tab
Dxvk Tab:
dxvk-rework-dxvk-tab
Version Switcher:
dxvk-rework-versions
Hud Switcher:
dxvk-rework-hud-options

### DIFF
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -82,7 +82,7 @@ public class CompatibilityTools
     }
 
         public CompatibilityTools(WineSettings wineSettings, Dxvk.DxvkHudType hudType, bool? gamemodeOn, bool? dxvkAsyncOn, DirectoryInfo toolsFolder)
-            : this(wineSettings, new DxvkSettings(hudType, "", dxvkAsyncOn), gamemodeOn, toolsFolder)
+            : this(wineSettings, new DxvkSettings(hudType, "", "", dxvkAsyncOn), gamemodeOn, toolsFolder)
         {
             // Old constructor format. This is for compatibility with XL.Core. Once changes are made
             // there, this can be deleted.
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -153,7 +153,7 @@ public class CompatibilityTools
 
         var wineEnviromentVariables = new Dictionary<string, string>();
         wineEnviromentVariables.Add("WINEPREFIX", Settings.Prefix.FullName);
-        wineEnviromentVariables.Add("WINEDLLOVERRIDES", $"mscoree=n;d3d9,d3d11,d3d10core,dxgi={(wineD3D ? "b" : "n")}");
+        wineEnviromentVariables.Add("WINEDLLOVERRIDES", $"msquic=,mscoree=n,b;d3d9,d3d11,d3d10core,dxgi={(DxvkSettings.Enabled && !wineD3D ? "n" : "b")}");
 
         if (!string.IsNullOrEmpty(Settings.DebugVars))
         {
@@ -163,15 +163,14 @@ public class CompatibilityTools
         wineEnviromentVariables.Add("XL_WINEONLINUX", "true");
         string ldPreload = Environment.GetEnvironmentVariable("LD_PRELOAD") ?? "";
 
-        if (this.gamemodeOn == true && !ldPreload.Contains("libgamemodeauto.so.0"))
+        if (gamemodeOn && !ldPreload.Contains("libgamemodeauto.so.0"))
         {
             ldPreload = ldPreload.Equals("") ? "libgamemodeauto.so.0" : ldPreload + ":libgamemodeauto.so.0";
         }
 
-        foreach(KeyValuePair<string,string> dxvkVar in DxvkSettings.DxvkVars)
-        {
+        foreach (KeyValuePair<string, string> dxvkVar in DxvkSettings.DxvkVars)
             wineEnviromentVariables.Add(dxvkVar.Key, dxvkVar.Value);
-        }
+
         wineEnviromentVariables.Add("WINEESYNC", Settings.EsyncOn);
         wineEnviromentVariables.Add("WINEFSYNC", Settings.FsyncOn);
 
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -168,7 +168,7 @@ public class CompatibilityTools
             ldPreload = ldPreload.Equals("") ? "libgamemodeauto.so.0" : ldPreload + ":libgamemodeauto.so.0";
         }
 
-        foreach (KeyValuePair<string, string> dxvkVar in DxvkSettings.DxvkVars)
+        foreach (var dxvkVar in DxvkSettings.DxvkVars)
             wineEnviromentVariables.Add(dxvkVar.Key, dxvkVar.Value);
 
         wineEnviromentVariables.Add("WINEESYNC", Settings.EsyncOn);
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -82,7 +82,7 @@ public class CompatibilityTools
     }
 
         public CompatibilityTools(WineSettings wineSettings, Dxvk.DxvkHudType hudType, bool? gamemodeOn, bool? dxvkAsyncOn, DirectoryInfo toolsFolder)
-            : this(wineSettings, new DxvkSettings(hudType, dxvkAsyncOn), gamemodeOn, toolsFolder)
+            : this(wineSettings, new DxvkSettings(hudType, "", dxvkAsyncOn), gamemodeOn, toolsFolder)
         {
             // Old constructor format. This is for compatibility with XL.Core. Once changes are made
             // there, this can be deleted.
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -65,28 +65,7 @@ public class CompatibilityTools
             if (!this.dxvkDirectory.Exists)
                 this.dxvkDirectory.Create();
         }
-
-        if (!wineSettings.Prefix.Exists)
-            wineSettings.Prefix.Create();
-        if (wineSettings.StartupType == WineStartupType.Managed)
-        {
-            if (!this.toolDirectory.Exists)
-                this.toolDirectory.Create();
-
-            if (!this.dxvkDirectory.Exists)
-                this.dxvkDirectory.Create();
-        }
-
-        if (!wineSettings.Prefix.Exists)
-            wineSettings.Prefix.Create();
     }
-
-        public CompatibilityTools(WineSettings wineSettings, Dxvk.DxvkHudType hudType, bool? gamemodeOn, bool? dxvkAsyncOn, DirectoryInfo toolsFolder)
-            : this(wineSettings, new DxvkSettings(hudType, "", "", dxvkAsyncOn), gamemodeOn, toolsFolder)
-        {
-            // Old constructor format. This is for compatibility with XL.Core. Once changes are made
-            // there, this can be deleted.
-        }
 
     public async Task EnsureTool(DirectoryInfo tempPath)
     {
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -43,17 +43,15 @@ public class CompatibilityTools
 
     public bool IsToolDownloaded => File.Exists(Wine64Path) && Settings.Prefix.Exists;
 
-    private readonly Dxvk.DxvkHudType hudType;
-    private readonly bool gamemodeOn;
-    private readonly string dxvkAsyncOn;
+    public DxvkSettings DxvkSettings { get; private set; }
 
-    public CompatibilityTools(WineSettings wineSettings, Dxvk.DxvkHudType hudType, bool? gamemodeOn, bool? dxvkAsyncOn, DirectoryInfo toolsFolder)
+    private readonly bool gamemodeOn;
+
+    public CompatibilityTools(WineSettings wineSettings, DxvkSettings dxvkSettings, bool? gamemodeOn, DirectoryInfo toolsFolder)
     {
         this.Settings = wineSettings;
-        this.hudType = hudType;
+        this.DxvkSettings = dxvkSettings;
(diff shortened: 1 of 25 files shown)


src/XIVLauncher.Common/Game/Launcher.cs
```diff
@@ -40,7 +40,7 @@ public class Launcher
 
     private const string FALLBACK_FRONTIER_URL_TEMPLATE = "https://launcher.finalfantasyxiv.com/v620/index.html?rc_lang={0}&time={1}";
 
-    public Launcher(ISteam? steam, IUniqueIdCache uniqueIdCache, ISettings settings, string? frontierUrl)
+    public Launcher(ISteam? steam, IUniqueIdCache uniqueIdCache, ISettings settings, string? frontierUrl = null)
```
marzent, line 43: This seems out of scope for this PR

src/XIVLauncher.Common/Game/Launcher.cs
```diff
@@ -71,7 +71,7 @@ public Launcher(ISteam? steam, IUniqueIdCache uniqueIdCache, ISettings settings,
         this.client = new HttpClient(handler);
     }
 
-    public Launcher(byte[] steamTicket, IUniqueIdCache uniqueIdCache, ISettings settings, string? frontierUrl)
+    public Launcher(byte[] steamTicket, IUniqueIdCache uniqueIdCache, ISettings settings, string? frontierUrl = null)
```
marzent, line 74: Ditto

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
```diff
@@ -0,0 +1,119 @@
...
+        frameRate ??= 0;
+        if (frameRate > 0) this.DxvkVars.Add("DXVK_FRAME_RATE", (frameRate).ToString());
+        this.DxvkVersion = version;
+        string release = this.DxvkVersion switch
+        {
+            Dxvk.DxvkVersion.v1_10_1 => "1.10.1",
+            Dxvk.DxvkVersion.v1_10_2 => "1.10.2",
+            Dxvk.DxvkVersion.v1_10_3 => "1.10.3",
+            Dxvk.DxvkVersion.v2_0 => "2.0",
+            _ => VersionOutOfRange(),
```
marzent, line 39: This probably should be an `ArgumentOutOfRangeException`, as it is not really supposed to happen and you'd have to change the default DXVK version in the future in like 3 different places

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
```diff
@@ -0,0 +1,119 @@
...
+                DxvkVars.Add("MANGOHUD","1");
+                DxvkVars.Add("MANGOHUD_CONFIG","full");
+                break;
+            // If DxvkHudType is None, or undefined, don't set anything.
+            default:
+                break;
+        }
+    }
+
+    private string VersionOutOfRange() {
```
marzent, line 96: This is unneeded with the above change

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
```diff
@@ -0,0 +1,119 @@
...
+    public string FolderName { get; private set; }
+
+    public Dictionary<string, string> DxvkVars { get; private set; }
+
+    public Dxvk.DxvkHudType DxvkHud { get; private set; }
+
+    public Dxvk.DxvkVersion DxvkVersion { get; private set; }
+
+    private const string ALLOWED_CHARS = "^[0-9a-zA-Z,=.]+$";
+    private const string ALLOWED_WORDS = "^(?:devinfo|fps|frametimes|submissions|drawcalls|pipelines|descriptors|memory|gpuload|version|api|cs|compiler|samplers|scale=(?:[0-9]){0,2}(?:.(?:[0-9])+)?)$";
```
marzent, line 21: Is there a reason for scale being < 100 ?
DXVK will allow such values and will sanitize them by itself, in fact the only restriction it enforces is scale >= 0.01:
https://github.com/doitsujin/dxvk/blob/c0d843c5788448ed66d61246c6ad8453c41c87fd/src/dxvk/hud/dxvk_hud.cpp#L13

src/XIVLauncher.Common.Unix/Compatibility/Dxvk.cs
```diff
@@ -49,7 +47,34 @@ public enum DxvkHudType
...
+
+    public enum DxvkVersion
+    {
+        [SettingsDescription("1.10.1 (default)", "The default version of DXVK used with XIVLauncher.Core.")]
+        v1_10_1,
+
+        [SettingsDescription("1.10.2", "Newer version of 1.10 branch of DXVK. Probably works.")]
+        v1_10_2,
+
+        [SettingsDescription("1.10.3", "Newer version of 1.10 branch of DXVK. Probably works.")]
```
marzent, line 74: 1.10.3 for sure works with Dalamud + GShade; it's really only 2.0 that is creating trouble

marzent (changes requested):
Thanks for bringing all DXVK settings together in one class, this is a nice change.

Just a few thoughts on even improving DXVK handling further:
Setting `DXVK_LOG_PATH` to the xlcore provided logs folder would be nice.
In the long run it would also be good to set `DXVK_CONFIG_FILE` and `DXVK_STATE_CACHE_PATH` to outside of the wine prefix, as that would make the shader state cache and DXVK configuration survive a prefix swap or deletion

src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
```diff
@@ -68,10 +66,28 @@ public CompatibilityTools(WineSettings wineSettings, Dxvk.DxvkHudType hudType, b
                 this.dxvkDirectory.Create();
         }
 
+        if (!wineSettings.Prefix.Exists)
```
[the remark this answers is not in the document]
rankynbass: So it is... deleting that. You know how it goes with copy/paste

src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
```diff
@@ -68,10 +66,28 @@ public CompatibilityTools(WineSettings wineSettings, Dxvk.DxvkHudType hudType, b
...
+
+            if (!this.dxvkDirectory.Exists)
+                this.dxvkDirectory.Create();
+        }
+
         if (!wineSettings.Prefix.Exists)
             wineSettings.Prefix.Create();
     }
 
+        public CompatibilityTools(WineSettings wineSettings, Dxvk.DxvkHudType hudType, bool? gamemodeOn, bool? dxvkAsyncOn, DirectoryInfo toolsFolder)
```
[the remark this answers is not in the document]
rankynbass: That's fine. I was just putting it in so it would still build against unmodified xlcore. Deleting this as  well.

src/XIVLauncher.Common/Game/Launcher.cs
rankynbass: Fair enough. It was just in so I could build against unmodified xlcore. Can definitely remove.

src/XIVLauncher.Common/Game/Launcher.cs
rankynbass: As above.

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
rankynbass: I didn't know if there were any limits, it just seemed a reasonable high number. Scales over 10 are probably not useful unless you're on an 8K+ display, but it's easy enough to change {0,2} to *. So... changed.

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
rankynbass: Alright, found a couple of examples in the code and mimicked them. It compiles, and, as you said, this shouldn't ever actually happen to a normal user. Changed.

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
rankynbass: And deleted.

rankynbass:
Added the dxvk log to logs folder, and added a .dxvk folder for state cache. It'll even be possible to pass in the correct storage path, assuming we ever get around to allowing changing that. By default, just assumes ~/.xlcore, though.

rankynbass:
Alright, should be good now as of commit df1a3be

rankynbass:
And I'm not sure if sync/async need different caches, so I'm keeping them separate now. Hopefully that's everything.

marzent:
> And I'm not sure if sync/async need different caches, so I'm keeping them separate now. Hopefully that's everything.

Theoretically they shouldn't, but there is some bugginess with async and DXVK 2.0 where async will miss saving entries sometimes, so this is probably cleaner

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
```diff
@@ -0,0 +1,121 @@
...
+    private const string ALLOWED_WORDS = "^(?:devinfo|fps|frametimes|submissions|drawcalls|pipelines|descriptors|memory|gpuload|version|api|cs|compiler|samplers|scale=(?:[0-9])*(?:.(?:[0-9])+)?)$";
+
+    public DxvkSettings(Dxvk.DxvkHudType hud = Dxvk.DxvkHudType.None, string dxvkHudCustom="",
+        string mangoHudPath="", bool? async = true, int? frameRate = 0,
+        Dxvk.DxvkVersion version = Dxvk.DxvkVersion.v1_10_1, string? corePath = null)
+    {
+        this.DxvkHud = hud;
+        string home = Environment.GetEnvironmentVariable("HOME");
+        corePath ??= Path.Combine(home,".xlcore");
+        string dxvkConfigPath = Path.Combine(corePath,".dxvk");
```
marzent, line 30: Maybe it is cleaner to pass a DirectoryInfo into the cctor here as well, just in case the base xlcore storage path will change

rankynbass: That's just the default. You can pass in the correct path to the constructor; it just defaults to null and sets the default to ~/.xlcore in that case. XL Core's Program.cs creates a DxvkSettings object, and I've passed it in there. But I could change it to pass in a DirectoryInfo. It's probably more "correct". So now the storage.Root object can be passed in from XL Core, instead of the string storage.Root.FullName.

rankynbass:
Pulled in marzent's commits. I am using the public static functions I declared in the XL.Core UI, so I added those back in.

rankynbass:
Got rid of a useless function that marzent previously removed and I foolishly re-added because I panicked when I saw a compile error about missing function `CheckMangoHudPath()`. Since all the function did was call `File.Exists()`, I just had to update XL.Core to use that instead of the function in question.

That should be it for now. Everything seems to be working when I build [PR #17](https://github.com/goatcorp/XIVLauncher.Core/pull/17), both natively and as a flatpak.

rankynbass:
Should we bump the default version to 1.10.3, and leave 1.10.1 and .2 in there for compatibility/testing purposes?

marzent:
> Should we bump the default version to 1.10.3, and leave 1.10.1 and .2 in there for compatibility/testing purposes?

1.10.3 is perfectly fine on any platform and plays nicer with some nvidia drivers iirc

marzent (approved):

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
```diff
@@ -0,0 +1,142 @@
...
+            FolderName = $"dxvk-async-{release}";
+            DxvkVars.Add("DXVK_ASYNC", async ? "1" : "0");
+        }
+        else
+        {
+            DownloadURL = $"https://github.com/doitsujin/dxvk/releases/download/v{release}/dxvk-{release}.tar.gz";
+            FolderName = $"dxvk-{release}";
+        }
+
+        DirectoryInfo dxvkCachePath = new DirectoryInfo(Path.Combine(dxvkConfigPath.FullName, "cache"));
```
marzent, line 65: Very minor nitpick, this is a bit inconsistent with the use of var everywhere else

src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
```diff
@@ -168,21 +163,14 @@ private Process RunInPrefix(ProcessStartInfo psi, string workingDirectory, IDict
...
-
-        if (this.gamemodeOn == true && !ldPreload.Contains("libgamemodeauto.so.0"))
+        if (gamemodeOn && !ldPreload.Contains("libgamemodeauto.so.0"))
         {
             ldPreload = ldPreload.Equals("") ? "libgamemodeauto.so.0" : ldPreload + ":libgamemodeauto.so.0";
         }
 
-        wineEnviromentVariables.Add("DXVK_HUD", dxvkHud);
-        wineEnviromentVariables.Add("DXVK_ASYNC", dxvkAsyncOn);
+        foreach (KeyValuePair<string, string> dxvkVar in DxvkSettings.DxvkVars)
```
marzent, line 171: same thing here

rankynbass: Alright, changed those to "var".

src/XIVLauncher.Common.Unix/Compatibility/DxvkSettings.cs
rankynbass: Changed.

rankynbass:
Made obsolete by #1366

Issue #12, opened 2019-03-11
Only able to log in if 'save for next startup' option is selected


BARASOAP:
I received an error when attempting to log in. The error happens regardless of whether or not I entered a valid password so I figured there may be a configuration error in my settings. **I eventually tried the 'save for next startup' option which then let the game launch**. This function may be broken.

Hopefully this isn't some dumb configuration thing I did and this info helps you at least a little

**Settings i used to resolve issue:**
![settings](https://user-images.githubusercontent.com/9269119/54149886-9b2a1200-440d-11e9-8620-e580467ec846.png)

**Screenshot of error:**
![error](https://user-images.githubusercontent.com/9269119/54150041-ee9c6000-440d-11e9-94f7-7efd53a42284.png)

**Error Text:**
```
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.ComponentModel.Win32Exception (0x80004005): 'CredDelete' call throw an error (Error code: 1168)
   at AdysTech.CredentialManager.CredentialManager.RemoveCredentials(String target)
   at XIVLauncher.MainForm.LoginButton_Click(Object sender, EventArgs e) in C:\projects\ffxivquicklauncher\XIVLauncher\MainForm.cs:line 106
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
   at System.Windows.Forms.Button.PerformClick()
   at System.Windows.Forms.Form.ProcessDialogKey(Keys keyData)
   at System.Windows.Forms.TextBoxBase.ProcessDialogKey(Keys keyData)
   at System.Windows.Forms.Control.PreProcessMessage(Message& msg)
   at System.Windows.Forms.Control.PreProcessControlMessageInternal(Control target, Message& msg)
   at System.Windows.Forms.Application.ThreadContext.PreTranslateMessage(MSG& msg)


************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.3324.0 built by: NET472REL1LAST_C
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll
----------------------------------------
XIVLauncher
    Assembly Version: 1.0.0.0
    Win32 Version: 1.0.0.0
[...]

PR #1206, opened 2022-12-10
fix: allow xlcore to build

rankynbass:
XL Core will not currently build. Need to update either the LauncherApp.cs in XL Core (see [PR #12](https://github.com/goatcorp/XIVLauncher.Core/pull/12)) or this. Both together also work.
* A 4th arg was added to constructor of Launcher class
* XL Core only knows about 3 args. Setting default to null allows build.

This was part of my dxvk settings rework, but was removed from there for being out of scope.


NotNite:
We have a separate repo for xlcore now, we should probably remove the one in here

PR #1209, opened 2022-12-11
Conversion to `System.Text.Json`

marzent:
Completely removes and replaces Newtonsoft.Json 

Everything should be functionally equivalent, except for `Headlines` deserialization, which works, but is not a 1:1 translation of the old options.

The remote patcher RPC has been slightly reworked to allow for deserialization without type information as well.


marzent:
Quickly merged master into this, but haven't tested the new version; is there any interest in this in general though?

goaaats:
Technically yes, but it's a bit hard to justify since there could be a lot of subtle bugs here and no huge benefits. You should leave this open as a memorial to what could be, if I have the time some day.

marzent:
Yeah you are probably right, this does make more sense on modern .NET anyways

PR #1212, opened 2022-12-13
Fix null inheritable process handle

marzent:
Fixes https://github.com/goatcorp/FFXIVQuickLauncher/issues/1211 (although would be interesting to know why it fails in the first place)


goaaats:
Probably some really cursed AV

PR #1218, opened 2023-01-01
Error icon URL causes long update time

AinaSnow:
A non-existent address in the squirrel pack causes long update times
NANOZPMY%F_8}A{87$3_(6X


goaaats:
Nice find, thanks. I'll probably host the icon off GitHub.

goaaats (approved):

PR #1242, opened 2023-01-15
Unify Dalamud injector arguments

BitsOfAByte:
This PR resolves the issues with the `UNIX` Dalamud injector arguments being very out of date by moving both the `Windows` and `UNIX` arguments into one class.

https://github.com/goatcorp/XIVLauncher.Core/pull/28 is blocked by this PR

CC: @NotNite

### DIFF
--- a/src/XIVLauncher.Common/Dalamud/DalamudInjectorArgs.cs
+++ b/src/XIVLauncher.Common/Dalamud/DalamudInjectorArgs.cs
@@ -1,0 +1,22 @@
+namespace XIVLauncher.Common.Dalamud
+{
+    public static class DalamudInjectorArgs
+    {
+        public const string Launch = "launch";
+        public const string WithoutDalamud = "--without-dalamud";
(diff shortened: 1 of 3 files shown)


src/XIVLauncher.Common/Dalamud/DalamudInjectorArgs.cs
```diff
@@ -0,0 +1,22 @@
+namespace XIVLauncher.Common.Dalamud
+{
+    public static class DalamudInjectorArgs
+    {
+        public const string Launch = "launch";
```
NotNite, line 5: These consts don't match our style guide, but I think we should focus on fixing this later with the Great Goatcorp Lint Bonanza.

NotNite (approved):
Builds fine on Windows 10 via Rider (but required a restart of Rider for intellisense to not die).

goaaats (approved):

PR #1266, opened 2023-02-07
[xlcore][feature] Add support for Proton as a runner

rankynbass:
This PR adds the necessary plumbing to use Proton as a runner for launching FFXIV. There are several reasons this is useful.
1. Steam Deck users will automatically have at least one Proton instance, and it will be designed to work with the Steam deck's unique input system.
2. Proton seems to have better support for non-latin character sets. It's possible to get this working with regular wine, but it's a bit tricky.
3. Proton updates fairly frequently, and also has experimental versions and GE-Proton with extra fixes. Since the XIVLauncher.Core changes will just read from steam's directories to find installed proton versions, we don't have to worry about keeping wine up to date. Valve (and GloriousEggroll) will do it for us.
4. This applies to DXVK (and possibly VKD3D in the future) as well. Proton installs maintain their own versions, so we don't have to worry about it. Older GE-Proton also includes dxvk async patches, and the newer ones use DXVK 2.1, which doesn't really need it.
5. Haptic feedback from PS5 controller works with a patched version of proton, but not (easily) with win. It's a nice-to-have feature.

This patch works primarily by modifying the RunInPrefix functions to use "proton runinprefix <command>" instead of "wine <command>" when proton is used. To get around the problem of needing a patched wine, I created a new function to get the unix pid by name. This should always work, unless you manage to launch a second copy of ffxiv_dx11.exe at exactly the right moment. The rest of the changes are just accommidating the proton and steam paths.

This PR is designed to work with a complimentary patch on the XL.Core side. However, if the current 1.0.3.0 is built against this as a submodule, Proton will appear as an option in the Wine Installation Type and will automatically select Proton 7.0. It assumes that Proton 7.0 is installed, and that steam is installed at the usual location of `~/.steam/root`.

Also, ignore the first commit comment about an environment variable and Dalamud not working. The environment variable was removed with the changes to WineSettings, and once I figured out how to get the unix pid by name, Dalamud worked.

### DIFF
--- a/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
+++ b/src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
@@ -41,11 +41,19 @@ public class CompatibilityTools
     private string Wine64Path => Path.Combine(WineBinPath, "wine64");
     private string WineServerPath => Path.Combine(WineBinPath, "wineserver");
 
+    private string SteamRoot => Path.Combine(Environment.GetEnvironmentVariable("HOME"),".steam","root");
+    private string ProtonPath => Path.Combine(SteamRoot,"steamapps","common","Proton 7.0","proton");
+
+    private string ProtonDataPath => Path.Combine(Settings.Prefix.Parent.FullName, "protonprefix");
+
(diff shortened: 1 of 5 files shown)


rankynbass:
Closing this as it is buggy. Running soldier runtime outside of steam is currently very unstable, and without that, Proton isn't really a good idea to use.

PR #1267, opened 2023-02-07
Remove a special exception for certain folders during repair

reiichi001:


goaaats:
Kizer already did this in his symlink fix, thanks though

PR #1278, opened 2023-02-12
Add reshade shader/presets folder to repair exemptions

Soreepeong:


reiichi001:
ReShade by default doesn't make a presets folder. It also doesn't make symlinks to anything like GShade did. We probably don't need to make any special exceptions for *shade anymore, as even GShade stopped making a gshade-shaders symlink as of 4.1.0, as that was moved to its Program Files folder.

I opened <https://github.com/goatcorp/FFXIVQuickLauncher/pull/1267> a little bit back to remove the exceptions / clean up some wording for integrity, since we'd still want to ignore them there for speed.

PR #1331, opened 2023-05-27
Always Log OTP Listener

reiichi001:
This is for troubleshooting purposes, as some users report that it's not starting / doesn't receive, and this will always ensure we know it started.

Just sets the debug log to info.


goaaats:
This won't help you at all, since the actual start is done in a task. You should put it inside the task, or better yet, into the continuation, where you can also observe the exception property if !IsCompletedSuccessfully.

reiichi001:
Does this look more in line with how it should be? Not used to using continuations, but still want the PR to become mergeable eventually.

PR #1348, opened 2023-06-24
Compatibility rework

rankynbass:
Here's the compatibilitytools rework I mentioned a few weeks ago. This moves most of the bits that will change over to the XL.Core side, such as specific versions of wine or dxvk, where they will be stored, and what environment variables need to be set for each one.

I also moved the MacVideoFix over to XL.Core. I figure individual implementations of fixes can be done on the XL.Core side, and then passed to GameFixApply object.


Blooym (approved):

src/XIVLauncher.Common.Unix/Compatibility/WineRunner.cs
```diff
@@ -0,0 +1,81 @@
...
+        Environment = (env is null) ? new Dictionary<string, string>() : env;
+        if (string.IsNullOrEmpty(runCmd))
+        {
+            RunCommand = Path.Combine(toolFolder, folder, "bin", (is64only) ? "wine" : "wine64");
+            WineServer = Path.Combine(toolFolder, folder, "bin", "wineserver");
+        }
+        else
+        {
+            if (File.Exists(Path.Combine(runCmd, "wine64")))
+                RunCommand = Path.Combine(runCmd, "wine64");
```
marzent, line 50: Is runCmd supposed to be a folder that the wine binaries reside in?
This is very confusing when having folder and rootFolder as well here.

src/XIVLauncher.Common.Unix/Compatibility/WineRunner.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    public string RunCommand { get; }
+
+    public string RunArguments { get; }
+
+    public string WineServer { get; }
+
+    public string PathArguments { get; }
+
+    private string wineFolder { get; }
```
marzent, line 24: why are some of these properties, and some fields

src/XIVLauncher.Common.Unix/Compatibility/WineRunner.cs
```diff
@@ -0,0 +1,81 @@
...
+                RunCommand = Path.Combine(runCmd, "wine");
+            else
+                throw new FileNotFoundException($"There is no wine or wine64 binary at {runCmd}.");
+            WineServer = Path.Combine(runCmd, "wineserver");
+        }
+    }
+
+    public async Task Install()
+    {
+        if (CompatibilityTools.IsDirectoryEmpty(Path.Combine(toolFolder, wineFolder)))
```
marzent, line 61: The wine "runner" should have a concept of it is pointing to custom user provided wine or is internally managed and guard against that here.

src/XIVLauncher.Common.Unix/Compatibility/WineRunner.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Serilog;
+using XIVLauncher.Common.Util;
+
+namespace XIVLauncher.Common.Unix.Compatibility;
+
+public class WineRunner
```
marzent, line 14: Not sure if this needs to be called runner, when all it does is hold wine settings/configuration

src/XIVLauncher.Common.Unix/Compatibility/WineRunner.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    private string downloadUrl;
+
+    private string toolFolder;
+
+    private string xlcoreFolder;
+
+    public Dictionary<string, string> Environment { get; }
+
+    public WineRunner(string runCmd, string runArgs, string folder, string url, string rootFolder, Dictionary<string, string> env = null, bool is64only = false)
```
marzent, line 34: `is64only` is a misnomer here, in general all wine builds with 64 bit support will have a dedicated `wine64` executable which is the one that should be invoked for xiv and Dalamud,  as when you use the wrong 32 bit one, wine will notice and call wine64 for you (except for some IL-only executables).

`is64only` should always be true, if anything, but is not even needed to be specified at all... The only exception to not calling wine64 is using the new experimental wine8 wow64 backend, which will create a single 64 bit only `wine` executable, that internally thunks all system calls when passing it a 32 bit PE executable (which ironically means when you have a true 64 bit only wine 8 `is64only` should be false here).

This is easy enough to handle transparently handle though, always pick `wine64` unless it doesn't exist, then pick `wine`.

src/XIVLauncher.Common.Unix/UnixDalamudRunner.cs
```diff
@@ -74,43 +74,57 @@ public UnixDalamudRunner(CompatibilityTools compatibility, DirectoryInfo dotnetR
...
         var dalamudProcess = compatibility.RunInPrefix(string.Join(" ", launchArguments), environment: environment, redirectOutput: true, writeLog: true);
-        var output = dalamudProcess.StandardOutput.ReadLine();
+        var output = dalamudProcess.StandardOutput.ReadLine() ?? "";
 
-        if (output == null)
-            throw new DalamudRunnerException("An internal Dalamud error has occured");
+        Console.WriteLine("[DALAMUD] " + output);
+
+        // Skip "ERROR: Could Not Get Primary Adapter Handle" and proceed to next line
+        if (output.Equals("ERROR: Could Not Get Primary Adapter Handle"))
```
marzent, line 82: Why is this necessary?
Wine should always output its internal logging to stderr, but this doesn't even look like wine.

src/XIVLauncher.Common.Unix/UnixDalamudRunner.cs
```diff
@@ -74,43 +74,57 @@ public UnixDalamudRunner(CompatibilityTools compatibility, DirectoryInfo dotnetR
         launchArguments.Add(gameArgs);
 
         var dalamudProcess = compatibility.RunInPrefix(string.Join(" ", launchArguments), environment: environment, redirectOutput: true, writeLog: true);
-        var output = dalamudProcess.StandardOutput.ReadLine();
+        var output = dalamudProcess.StandardOutput.ReadLine() ?? "";
```
marzent, line 77: `output` is checked against null later, no need to null-coalesce it twice.
Also `output` being null is meaningful information, since it indicates that the process exited without writing a single line to stdout

src/XIVLauncher.Common.Unix/UnixDalamudRunner.cs
```diff
@@ -74,43 +74,57 @@ public UnixDalamudRunner(CompatibilityTools compatibility, DirectoryInfo dotnetR
...
             }
 
         }).Start();
 
+        int unixPid = 0;
+        int winePid = 0;
         try
         {
             var dalamudConsoleOutput = JsonConvert.DeserializeObject<DalamudConsoleOutput>(output);
-            var unixPid = compatibility.GetUnixProcessId(dalamudConsoleOutput.Pid);
```
marzent, line 98: All of `UnixDalamudRunner` can be left as is with the adjustments to `GetUnixProcessId` from before

marzent (changes requested):

src/XIVLauncher.Common.Unix/Compatibility/CompatibilityTools.cs
```diff
@@ -17,97 +17,75 @@ namespace XIVLauncher.Common.Unix.Compatibility;
...
-    private const string WINE_XIV_RELEASE_URL = "https://github.com/goatcorp/wine-xiv-git/releases/download/8.5.r4.g4211bac7/wine-xiv-staging-fsync-git-fedora-8.5.r4.g4211bac7.tar.xz";
-#else
-    private const string WINE_XIV_RELEASE_URL = "https://github.com/goatcorp/wine-xiv-git/releases/download/8.5.r4.g4211bac7/wine-xiv-staging-fsync-git-ubuntu-8.5.r4.g4211bac7.tar.xz";
-#endif
-    private const string WINE_XIV_RELEASE_NAME = "wine-xiv-staging-fsync-git-8.5.r4.g4211bac7";
-
     public bool IsToolReady { get; private set; }
 
-    public WineSettings Settings { get; private set; }
+    public WineRunner WineSettings { get; private set; }
```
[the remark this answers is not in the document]
rankynbass: Good point. Can certainly change this back to WineSettings.

It made more sense when I started the whole thing, honestly. The names just kind of stuck around. I can certainly change it back. When I started this whole thing, Dxvk and Wine were both in the same object, but it became too complicated and janky, so I split them. I think Runner was chosen simply because that's the folder lutris stuffs wine installs into.

rankynbass:
I'll certainly check into all of this when I get a chance later today. Thanks for good look at this.

rankynbass:
First, marzent, thank you for your thorough review of this PR. It was very helpful.
Second, rather than spam everyone with 20 emails for responding to your feedback, I’ll address it all here (hopefully in order).
- Renamed WineRunner and DxvkRunner to WineSettings and DxvkSettings. I suppose I could have kept Dxvk instead of DxvkSettings, but I think this is better just for consistency.
- WineDLL overrides: yeah, it’s all over the place. I’ve fixed it to just do it in compatibilitytools now, by checking for DxvkSettings.IsDxvk and setting it appropriately. There is still an override for wineD3D in big RunInPrefix() function, but I think that should stay there. There might be a reason we would want to launch a process with wineD3D while keeping the game in Dxvk.
- Got rid of the extra special environment variable for the moment. Might bring it back later if it’s needed, but it wasn’t actually used.
- Renamed beta to wine in all cases I could find. It compiles and puts files in the right place, so I assume I got them all.
- Moved the two install functions from the two Settings classes back into CompatibilityTools. Those functions are now called in the EnsureTool() function. Renamed them to EnsureWine() and EnsureDxvk().
- The logfile entries with (string) and (array) were in there because I was trying to track down exactly which one was being called and causing me problems. This has been removed.
- The LD_PRELOAD merge is there because I was trying to get mangohud to work with opengl. This is actually the reason a lot of the extra variables are in there; I had to basically hijack the WineSettings.RunCommand, set it to mangohud, pass a bunch of arguments (including the original wine binary), and set an LD_PRELOAD value to make sure it got loaded properly. This happened in Dxvk.Environment, which could conflict with WineSettings.Environment and the gamemode setting. So the merge was to make sure they played nice with each other. I’ve left this in for now.
- I’ve added a public property to CompatibilityTools, and passed it through the constructor. It’s not used for anything yet, but might be useful in the future, and now it’s in the Common.Unix side. I commented out the FLATPAK_NOTRIGHTNOW stuff.
- I left the GetUnixProcessIdByName() as public, because I still want to call it in one specific case.
- Cleaned up a bunch of leftover variables from DxvkSettings and WineSettings. The runCmd variable in the WineSettings constructor has been renamed to customwine, to indicate that’s what it actually is used for.
- The is64only stuff was for Wine built with WoW64, which supports 32-bit windows programs without needing 32-bit linux libs. You’re absolutely right that this can be done transparently, though.
- I restored the UnixDalamudRunner.cs file, mostly. I added an extra call to GetUnixProcessIdByName() into the catch statement at the end, because sometimes Dalamud will through out an extra error line that can’t be parsed by the JsonReader because it’s obviously not json. But Dalamud often keeps running regardless, and we can still potentially find the game process. So I think it’s better to make a second attempt than to return null, since that can cause the XIVLauncher.Core process to stick around after FFXIV exits.

For reasons unknown, after doing all these changes the Wine 7.11 that I tested for you the other day started working. It still fails in the current flatpak and rpm builds (and on the previous test builds I posted), but with these changes it now properly loads Dalamud, as if there were no problems at all. Previously, the game would load, but the process would hang trying to read the output. With these changes it doesn’t, and I have no idea why.
 
I’ll look at the XL.Core side tomorrow.

Thanks!

rankynbass:
I made a mistake in detecting wine or wine64 binary. This was actually the cause of the Wine 7.11 build working: it actually wasn't. Instead, I was accidentally trying to find wine (not wine64), failing, and then falling back to whichever Managed wine option was last selected. So the 7.11 wine still doesn't work, and I've fixed the logic.

src/XIVLauncher.Common.Unix/UnixDalamudRunner.cs
```diff
@@ -74,43 +74,57 @@ public UnixDalamudRunner(CompatibilityTools compatibility, DirectoryInfo dotnetR
...
+        string output;
+        int dalamudErrorCount = 0;
+        do
+        {
+            output = dalamudProcess.StandardOutput.ReadLine();
+            if (output is null)
+                throw new DalamudRunnerException("An internal Dalamud error has occured");
+            Console.WriteLine("[DALAMUD] " + output);         
+            dalamudErrorCount++;
+        } while (!output.StartsWith('{') && dalamudErrorCount <= 2);
```
marzent, line 89: This is a very crude check for valid JSON we should try to parse lines until they can be deserialized to a `DalamudConsoleOutput`, something like https://github.com/marzent/FFXIVQuickLauncher/commit/f50acf410700c1c7ee02e8bd9705e5fe8dddb58c would be better IMO...
Although I generally dislike both of these solutions that issue should really not be happening in the first place, but I guess some configurations are special

src/XIVLauncher.Common.Unix/UnixDalamudRunner.cs
```diff
@@ -74,43 +74,57 @@ public UnixDalamudRunner(CompatibilityTools compatibility, DirectoryInfo dotnetR
...
 
         new Thread(() =>
         {
             while (!dalamudProcess.StandardOutput.EndOfStream)
             {
                 var output = dalamudProcess.StandardOutput.ReadLine();
                 if (output != null)
-                    Console.WriteLine(output);
+                    Console.WriteLine("[DALAMUD] " + output);
```
marzent, line 97: This is confusing since it really is Injector and not Dalamud proper here (in any case best to leave it be)

src/XIVLauncher.Common.Unix/UnixDalamudRunner.cs
marzent: If `GetUnixProcessId` is properly implemented all of this can be really left as as (save the JSON reading adjustment  and spelling mistakes) to keep it simple

rankynbass:
Closing and replacing with https://github.com/goatcorp/FFXIVQuickLauncher/pull/1366. There were too many changes needed to rework this branch directly, and it was better to just start over from current master.

PR #1353, opened 2023-06-30
added split button for launch button context menu

### LINKED ISSUE #1337: Login button should be a split button
...and offer the current context menu items, to indicate that there are context menu items to be had.

image

Silicasandwhich:
Added a separate, thin button next to the launch button that, when clicked, displays what it currently the context menu of the launch button. The style of this button is as close to the style of the other buttons as possible, but the drop shadow may be slightly different. Additionally, the context menu of the launch button is still present for the sake of not breaking current behavior

fixes #1337


goaaats:
Thanks, much appreciated!

PR #1354, opened 2023-07-03
fix: Update name of redistributable package to download

Dormanil:
Ever since Microsoft released a new version of Visual Studio, there has been a change to the name of the Visual C++ Redistributable. I think it is about time we updated our language, especially since people don't necessarily know that we are telling them outdated information.

This also renames the localisation key. If you don't want that, I can undo that particular change.

### DIFF
--- a/src/XIVLauncher/Resources/Loc/XIVLauncher_Localizable.json
+++ b/src/XIVLauncher/Resources/Loc/XIVLauncher_Localizable.json
@@ -567,8 +567,8 @@
     "message": "XIVLauncher is already patching your game in another instance. Please check if XIVLauncher is still open.",
     "description": "<RepairGame>d__25.MoveNext"
   },
-  "DalamudVc2019RedistError": {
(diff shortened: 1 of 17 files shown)


KazWolfe:
Looks like this may be an annoying pain, but it would be awesome if we could include a button/link to the file specifically. Is there any way we can hack that in? (at-maintainers)

reiichi001:
We already have a FAQ to cover it, when it does come up.

Otherwise, that one SquirrelSetup fork can technically check/enforce it, but we don't typically run into this very often given SquirrelSetup itself will also generally fail if the VC Redist is missing entirely. (This check really just ensures the user has one that covers everything Dalamud needs instead of some ancient install of just vcrun2015 or something)

It's possible to add a link or button on the error popup message if they made it past Squirrel, but I'm not sure if we entirely need it. <https://aka.ms/vs/17/release/vc_redist.x64.exe> is available as a permalink from Microsoft, as the runtime installer hasn't changed since 2015.

Dormanil:
That link is nice and good, but it's not always been the same link ever since 2015. Replacing the `17` with a `15` results in the 2017 installer, and replacing it with a `16` results in the 2015-2019 one.

PR #1355, opened 2023-07-09
Update dalamudAsset cleanup to scan for all directories not current or dev

belthesar:
Per discussion in #xl-dev, this change updates the cleanup routine used to clean up old dalamudAsset versions to scan through the entire directory tree instead of the last 30 or so.

### DIFF
--- a/src/XIVLauncher.Common/Dalamud/AssetManager.cs
+++ b/src/XIVLauncher.Common/Dalamud/AssetManager.cs
@@ -166,7 +166,7 @@ namespace XIVLauncher.Common.Dalamud
 
             Log.Verbose("[DASSET] Assets OK at {0}", assetsDir.FullName);
 
-            CleanUpOld(baseDir, info.Version - 1);
+            CleanUpOld(baseDir, info.Version);
 
             return (assetsDir, info.Version);
         }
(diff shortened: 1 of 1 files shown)


src/XIVLauncher.Common/Dalamud/AssetManager.cs
```diff
@@ -225,16 +225,14 @@ private static void CleanUpOld(DirectoryInfo baseDir, int version)
...
 
-            for (int i = version; i >= version - 30; i--)
+            foreach (DirectoryInfo toDelete in baseDir.GetDirectories())
             {
-                var toDelete = Path.Combine(baseDir.FullName, i.ToString());
-
                 try
                 {
-                    if (Directory.Exists(toDelete))
+                    if (Directory.Exists(toDelete.FullName) && toDelete.Name != "dev" & toDelete.Name != version.ToString())
```
rankynbass, line 232: Pretty sure that should be a `&&` between `"dev"` and `toDelete.Name`, unless I misunderstand what you're doing here. Although I suppose a bitwise & of two boolean values might have the same result.

belthesar: Ahh beans, you're absolutely right. It definitely works as is, but that's not what is intended. Good catch, this was me not correcting the second `&` after the first not working.

PR #1378, opened 2023-09-08
Add YubiKey authentication functionality

ThatOneCamel:
Adds ability to use a YubiKey security key to generate OTP codes and use them to login.

Please also consider this corresponding pull request on the faq repo: [#tbd](https://github.com/goatcorp/faq/pull/tbd)
- This guide outlines what a YubiKey is and how a user would setup their own to work with XIVLauncher.

---
### Additions:

SettingsControl
- Adds a new CheckBox that enables usage of YubiKey hardware [this requires OTP macros to be enabled]
- Added way for users to setup their YubiKey credentials in-app (See: SecurityKeySetupDialog)

YubiAuth
 - Finds and maintains reference to physical YubiKey device
 - Handles most methods that retrieve information from the security key and perform operations

OtpInputDialog
- If YubiKey setting is enabled, will attempt to login if device is plugged in or, if applicable, prompt the user to touch their device before authentication occurs
- Made so security key authentication doesn't block the main thread and normal login can occur if user desires

SecurityKeySetupDialog
- A dialog that allows users to enter their authentication key and create a credential on their security key
- User can also decide if they want to use touch prior to authentication

Directory.Build.targets
- Made to ensure the proper DLL is selected for the proper system architecture
  - This was incorrectly imported using the default Yubico.YubiKey implementation

### DIFF
--- a/src/XIVLauncher/Accounts/YubiAuth.cs
+++ b/src/XIVLauncher/Accounts/YubiAuth.cs
@@ -1,0 +1,125 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Serilog;
+using Yubico.YubiKey;
+using Yubico.YubiKey.Oath;
+using Yubico.PlatformInterop;
+using System.Runtime.InteropServices;
+
+namespace XIVLauncher.Accounts
+{
+    public class YubiAuth
+    {
+        private const string ISSUER = "XIVLauncher";
+        private const string ACCOUNT_NAME = "FFXIV";
+        private const CredentialType AUTH_TYPE = CredentialType.Totp;
+        private const CredentialPeriod TIME_PERIOD = CredentialPeriod.Period30;
+        private const byte NUM_DIGITS = 6;
+
+        private static IYubiKeyDevice _yubiKey;
+        public YubiKeyDeviceListener DeviceListener;
+        public YubiAuth()
+        {
+            FindYubiDevice();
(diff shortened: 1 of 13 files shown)


KazWolfe:
Peanut gallery commentary (and not something necessary for this PR):

This would be a great opportunity to implement some kind of authentication plugin logic to XL so that people can load in all sorts of custom behavior (e.g. this, 1Password, the TOTP webserver, highly insecure homebaked DLLs with hardcoded TOTP secrets, etc.) that can provide credentials or TOTP to XL.

NotNite:
+1 for that idea @KazWolfe mentioned (but neither of us are XL team so don't take it as official word). Perhaps just exchanging messages over stdin/stdout in an executable you can configure sounds good.

KazWolfe:
And an actual comment:

At present, this does not appear to support multiple accounts. Would it be possible to tweak the behavior slightly to encode the SE username into the credential so that we can gracefully handle multi-service-account cases?

ThatOneCamel:
> 



> And an actual comment:
> 
> At present, this does not appear to support multiple accounts. Would it be possible to tweak the behavior slightly to encode the SE username into the credential so that we can gracefully handle multi-service-account cases?


> And an actual comment:
> 
> At present, this does not appear to support multiple accounts. Would it be possible to tweak the behavior slightly to encode the SE username into the credential so that we can gracefully handle multi-service-account cases?

Thanks for the feedback! Yes this is possible and I've gone ahead and tweaked the implementation so that this could happen.

PR #1426, opened 2023-12-18
Implement config import/export buttons for in game addon

reiichi001:
Adds buttons to import/export a zip file containing Dalamud and Plugin config. No launcher config is being handled here as this can be platform and computer specific.

Adds WPF commandline arguments to handle import/export as well for testing purposes that do not rely on gui features.

Import and Export are implemented under XIVLauncher.Common so that this can be implemented on other compatible launchers that use xlcommon.

### DIFF
--- a/src/XIVLauncher.Common/Support/DalamudAndPluginConfigImportExport.cs
+++ b/src/XIVLauncher.Common/Support/DalamudAndPluginConfigImportExport.cs
@@ -1,0 +1,121 @@
+using System;
+using System.Collections.Generic;
+using System.IO.Compression;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace XIVLauncher.Common.Support
+{
+    public class DalamudAndPluginConfigImportExport
+    {
+        public static async Task<string> ExportConfig(string storagePath)
+        {
+            // grab any platform-specific details
+            // there's none to grab currently :)
+
+            // start making our zip
+            var outFile = new FileInfo(Path.Combine(storagePath, $"XIVLauncher-{DateTime.Now:yyyyMMddhhmmss}.xlconf"));
+            using var archive = ZipFile.Open(outFile.FullName, ZipArchiveMode.Create);
+
+            // add all the stock XIVLauncher based paths
+            var accountsListFile = Path.Combine(storagePath, "accountsList.json");
+            var dalamudConfigFile = Path.Combine(storagePath, "dalamudConfig.json");
+            var dalamudVfsFile = Path.Combine(storagePath, "dalamudVfs.db");
+            var pluginConfigsFolder = Path.Combine(storagePath, "pluginConfigs");
+
+            AddIfExist(accountsListFile, archive);
+            AddIfExist(dalamudConfigFile, archive);
+            AddIfExist(dalamudVfsFile, archive);
+            AddIfExist(pluginConfigsFolder, archive);
+
+            // add some known special exceptions. It might be better to not build these expectations though
+            var backupsFolder = Path.Combine(storagePath, "backups"); // Otter plugins
+            var playerTrackBackupsFolder = Path.Combine(storagePath, "playerTrackBackups"); // PlayerTrack
+
+            AddIfExist(backupsFolder, archive);
+            AddIfExist(playerTrackBackupsFolder, archive);
+
+            // return the folder containing our exported settings
+            return outFile.FullName;
+        }
+
+        public static void ImportConfig(string zipFilePath, string storagePath)
+        {
+            // grab any platform-specific details
+            // there's none to grab currently :)
+
+            // TODO: Decide if we're going to alert on overwriting config.
+            // Right now, Franz decided against it. The user has to intentionally try to use this feature
+            // Also, .Net Framework is dumb and will explode if we use ZipArchive.ExtractToDirectory()
+            // and there are any file conflicts. .Net Core doesn't have this issue though and provides
+            // an override we could have used to do it anyways. Alternatively, we could just delete
+            // all of the files/folders we'd be restoring to first, but that also feels bad.
+
+
+
+            var inFile = new FileInfo(zipFilePath);
+            using var archive = ZipFile.Open(inFile.FullName, ZipArchiveMode.Read);
+
+            // If we weren't on .Net Framework, we could use this...
+            // ZipFileExtensions.ExtractToDirectory(archive, storagePath, true);
+
+            foreach (var entry in archive.Entries)
+            {
+                var extractPath = storagePath + "\\" + entry.FullName;
+                // If we were going to warn about overwriting files, it would go here.
+                /*
+                bool promptAlwaysForOverwrite = true;
+                if (promptAlwaysForOverwrite && File.Exists(extractPath))
+                {
+                    // Make some prompt. Overwrite? Yes, Yes to All, Cancel
+
+                    if (result == no) break or something.
+                    if (result == yestoall) promptAlwaysForOverwrite = false;
+                    // yes is the default and needs no special handling.
+                }
+                */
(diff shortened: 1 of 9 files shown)


reiichi001:
Sample screenshots of what the GUI implementation looks like.
image
image
image
image

src/XIVLauncher.Common/Support/DalamudAndPluginConfigImportExport.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            var inFile = new FileInfo(zipFilePath);
+            using var archive = ZipFile.Open(inFile.FullName, ZipArchiveMode.Read);
+
+            // If we weren't on .Net Framework, we could use this...
+            // ZipFileExtensions.ExtractToDirectory(archive, storagePath, true);
+
+            foreach (var entry in archive.Entries)
+            {
+                var extractPath = storagePath + "\\" + entry.FullName;
```
goaaats, line 64: Path.Combine()?

src/XIVLauncher.Common/Support/DalamudAndPluginConfigImportExport.cs
```diff
@@ -0,0 +1,121 @@
...
+            // an override we could have used to do it anyways. Alternatively, we could just delete
+            // all of the files/folders we'd be restoring to first, but that also feels bad.
+
+
+
+            var inFile = new FileInfo(zipFilePath);
+            using var archive = ZipFile.Open(inFile.FullName, ZipArchiveMode.Read);
+
+            // If we weren't on .Net Framework, we could use this...
+            // ZipFileExtensions.ExtractToDirectory(archive, storagePath, true);
```
goaaats, line 60: It's probably better to go through the entries directly. Then you can also choose to delete the file if it exists - which is probably what people expect when importing a backup.

src/XIVLauncher/App.xaml.cs
```diff
@@ -59,6 +60,12 @@ public class CmdLineOptions
             [CommandLine.Option("clientlang", Required = false, HelpText = "Client language to use.")]
             public ClientLanguage? ClientLanguage { get; set; }
 
+            [CommandLine.Option("exportconfig", Required = false, HelpText = "Export a zip of all config")]
```
goaaats, line 63: Are these necessary? What's your use case?

src/XIVLauncher/Windows/SettingsControl.xaml.cs
```diff
@@ -389,6 +390,84 @@ private void GamePathEntry_OnTextChanged(object sender, TextChangedEventArgs e)
...
+            {
+                try
+                {
+                    // Open document
+                    string zipFile = selectConfImportDialog.FileName;
+
+                    //TODO: Localize this text like a good programmer
+                    
+                    var importConfirmPromptResult = CustomMessageBox.Show(Loc.Localize("ConfigImportPrompt1", "XIVLauncher is going to import config from the file you selected.\n\n") + zipFile + Loc.Localize("ConfigImportPrompt2", "\n\nThis will overwrite any files that already exist.\n\nDo you want to proceed?"),
+                        Loc.Localize("ConfigImportPromptTitle", "Import confg file?"), MessageBoxButton.YesNo);
```
goaaats, line 414: Typo!

src/XIVLauncher/Windows/SettingsControl.xaml.cs
```diff
@@ -389,6 +390,84 @@ private void GamePathEntry_OnTextChanged(object sender, TextChangedEventArgs e)
...
+            }
+        }
+        private void ExportConfigButton_OnClick(object sender, RoutedEventArgs e)
+        {
+            try
+            {
+                var window = new ConfigImportExportProgressWindow();
+                window.SetTitle(Loc.Localize("ConfigExportRunning", "Running config export..."));
+
+                Task.Run(async () => await DalamudAndPluginConfigImportExport.ExportConfig(Paths.RoamingPath)).ContinueWith(async task =>
```
goaaats, line 450: This task needs a continuation. You need to observe the exception if it didn't complete successfully and hide the progress window if it's still open.

src/XIVLauncher/Windows/SettingsControl.xaml.cs
```diff
@@ -389,6 +390,84 @@ private void GamePathEntry_OnTextChanged(object sender, TextChangedEventArgs e)
...
+                    CustomMessageBox.Show(Loc.Localize("ConfigExportSuccess", "Exported config as ") + exportedFile.Name, "XIVLauncher", MessageBoxButton.OK, MessageBoxImage.Information);
+                    Process.Start("explorer.exe", $"/select, \"{filename}\"");
+                });               
+
+                window.ShowDialog();
+
+            }
+            catch (Exception ex)
+            {
+                CustomMessageBox.Show(ex.ToString(), "XIVLauncher Error", MessageBoxButton.OK, MessageBoxImage.Error);
```
goaaats, line 467: Hide the progress window here too?

src/XIVLauncher.Common/Support/DalamudAndPluginConfigImportExport.cs
```diff
@@ -0,0 +1,121 @@
...
+                */
+                if (!Directory.Exists(Path.GetDirectoryName(extractPath)))
+                {
+                    Directory.CreateDirectory(Path.GetDirectoryName(extractPath));
+                }
+                ZipFileExtensions.ExtractToFile(entry, extractPath, true);
+            }
+        }
+
+        private static void AddIfExist(string entryPath, ZipArchive zip)
```
goaaats, line 85: We already have this function in the tspack code. Maybe move to common util class?
https://github.com/goatcorp/FFXIVQuickLauncher/blob/6f388fb7787952e837b851b70dbec9ab8349eb3e/src/XIVLauncher/Support/PackGenerator.cs#L39

goaaats (changes requested):

src/XIVLauncher/App.xaml.cs
reiichi001: Not strictly needed. I just like having command args for any of the file-based features. Makes it easy for testing when I want to specify custom appdata paths to certify that it unpacks correctly.

src/XIVLauncher/Windows/SettingsControl.xaml.cs
reiichi001: `confg` spelling corrected. Was that the only one? Or did I miss another? (Will be included on next commit on this branch)

src/XIVLauncher.Common/Support/DalamudAndPluginConfigImportExport.cs
reiichi001: This one's a little different from tspack code. I had to get a little hacky for directory handling. But that could probably still be used for tspacks too. I've moved it to a util method in xivlauncher.common.util, but the WPF app didn't really like me doing that and VS stopped giving me intellisense when I tried. I might need some handholding to get that moved properly. (It could just be my machine being annoying)

src/XIVLauncher.Common/Support/DalamudAndPluginConfigImportExport.cs
reiichi001: My initial vision was to just overwrite from the backup, but ideally yes, a "do you want to replace this file, yes, yes to all, no, cancel" style prompt would be better. My worry here is some plugins generate TONS of files in their config directories and most users are not going to want to click yes to that every time. (ex: CharacterSync making backups of all game config for multiple characters for multiple days/launches)

This was more of a companion to the official launcher's FEA file solution, but for our config. (Although nothing prevents this from later handling game config too if we wanted to do that)

PR #1451, opened 2024-03-19
Fix some boot hash targets not existing anymore

NotNite:
They deleted some of the files we hash for boot so I'm just skipping the ones that don't exist. Also because the last entry isn't guaranteed to exist I moved to TrimEnd for the comma.

### DIFF
--- a/src/XIVLauncher.Common/Game/Launcher.cs
+++ b/src/XIVLauncher.Common/Game/Launcher.cs
@@ -360,14 +360,12 @@ public class Launcher
 
         for (var i = 0; i < FilesToHash.Length; i++)
         {
-            result +=
-                $"{FilesToHash[i]}/{GetFileHash(Path.Combine(gamePath.FullName, "boot", FilesToHash[i]))}";
-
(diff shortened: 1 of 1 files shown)


NotNite:
Also: not sure if `i` matters here? This passed boot hash checks but I don't know if it's 100% accurate boot behavior (not that we're doing accurate boot behavior anyways lmao).

NotNite:
Checked by Fiddler'ing boot, it matches current behavior.

marzent:
Wouldn't it be better to adjust https://github.com/goatcorp/FFXIVQuickLauncher/blob/f945283ea795e3ee0910b86b2e663971bd0abeaf/src/XIVLauncher.Common/Game/Launcher.cs#L80

NotNite:
They both have the same result, but I can do that if it's wanted. Gimme a few seconds...

marzent (approved):
Thanks IMO it is a bit clearer that way

avafloww (approved):

Blooym (approved):
Just here to give my grey checkmark

PR #1488, opened 2024-04-30
Bump fallback ZIP url

KazWolfe:
1.1.2 is dead, long live 1.1.4


Blooym (approved):
First time setup on Linux requires this fix as soon as possible

PR #1509, opened 2024-06-20
[CI/CD] Update action-gh-release

Bluefissure:
I think github deprecated some of the APIs so the action at that commit [failed as 404 on our end](https://github.com/ottercorp/FFXIVQuickLauncher/actions/runs/9557942271/job/26345998534):
image

Update it to a74c6b72af54cfa997e81df42d94703d6313a2d0 (action-gh-release@v2.0.6) to be compatible with the latest github release API.

Also chages the `\` in path to `/` as said in [README of action-gh-release](https://github.com/softprops/action-gh-release?tab=readme-ov-file#%EF%B8%8F-uploading-release-assets).

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -54,7 +54,7 @@ jobs:
           rm ".\Releases\$($delta_file.name)"
           rm ".\Releases\$($full_file.name)"
       - name: Create Release
-        uses: softprops/action-gh-release@v2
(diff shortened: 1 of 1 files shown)


goaaats:
Could you specify a commit hash instead of a tag? I prefer not to use tags for actions not provided by GitHub.

Bluefissure:
> Could you specify a commit hash instead of a tag? I prefer not to use tags for actions not provided by GitHub.

Sure, it's now using the commit hash from [action-gh-release@v2.0.6](https://github.com/softprops/action-gh-release/releases/tag/v2.0.6).

PR #1591, opened 2024-07-28
Fix xlcore oathloginexception

rankynbass:
Adds logic to OauthLoginException to deal with XL.Core error messages. There may be a way to do it by modifying the regex, but this works fine in my testing. I created a temporary env variable to set the login state manually. The errors will now show up with the error text instead of displaying "Unknown error".


rankynbass:
Alternately, to make it really simple, we could just change one line (28), to return the document string instead of null. Then any message that didn't match the regex would just get used as-is, which would include those from XL.Core.

rankynbass:
Closing this; can be fixed on XL.Core side.

PR #1629, opened 2024-11-12
Update App.xaml.cs

Archmrn:
remove unnecessary garbage

Why this even exists? Unreadable text, visual inconsistency


Archmrn:
So, why?

PR #1630, opened 2024-11-12
Update App.xaml.cs

Archmrn:
Why this even exists? Unreadable text, visual inconsistency.

#1629 Was closed without any comments, why? What's wrong with removing something that interferes with interface of app?

### DIFF
--- a/src/XIVLauncher/App.xaml.cs
+++ b/src/XIVLauncher/App.xaml.cs
@@ -95,12 +95,6 @@ namespace XIVLauncher
         public static byte[] GlobalSteamTicket { get; private set; }
         public static DalamudUpdater DalamudUpdater { get; private set; }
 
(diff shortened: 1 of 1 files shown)


Aireil:
You're right, the yellow part is quite hard to read, although I don't think your fix is ideal one.
It would be better if you just replaced the font color instead, either for all or just in the hard to read parts.
Thank you for your contribution!

Archmrn:
Hey

You mean that 'brush' thing is intended to be there no matter cause? 
Why don't just remove it, other languages doesnt trigger something like that

TeranFethril:
I think it is part of the process that Russia's unjustified war of aggression against Ukraine also reaches Russian-speaking users.
Political matters should definitely be kept out of this, even if it seems justified.

PR #1642, opened 2024-11-20
Dawntrail ready!

Aireil:
Update MaxExpansion on fake start


Aireil:
I don't even know if it matters now with the new title screen thing, had that changed locally for a while.
Well, it can't hurt.

PR #1650, opened 2024-12-01
Detect if running elevated

nathanctech:
Simple code (with command-line arg override) to detect if running elevated and toss an error to the user if so. Hopefully will prevent clobbered file permissions in the future. Tested on Windows 11 with a local administrator and UAC prompt. It does not trigger if UAC is disabled on the machine, though.


KazWolfe:
Ref goatcorp/Dalamud#1137

Issue #1433, opened 2024-01-30
[Proposal] Authentication Plugin System


KazWolfe:
We've received a number of requests/PRs for various interesting authentication methods (e.g. #1378, #1432, #738, and more), and this is more or less a repost of something posted on one of those Issues.

XIVLauncher may benefit from supporting plugins designed to allow integration with authentication providers. These plugins would provide an API surface for users to add support for their own custom authentication integrations however they see fit. 

The primary intent of the authentication plugin system would be to provide a deep connection between XIVLauncher and any given password manager or provider. The API surface should include functionality for:

* Providing a list of saved accounts from the backend store
* Providing the credentials (username + password + optionally OTP) from any given backend
* Retrieving an OTP code from a backend, allowing devs to write their own OTP providers

This would open a number of use cases, such as:

* Allowing XIVLauncher to grab all Square Enix accounts from 1Password, allowing the user to choose the saved password to use, and logging in with it.
* Allowing developers to store their (encrypted) credentials backed by a U2F security key rather than the Windows Credential Store.
* Allowing porting the OTP server feature to its own plugin.

The Authentication Plugin feature should be documented and open for developers to use, although the project can officially support certain "known-good" integrations (e.g. those for 1Password, Bitwarden, and others). A type of "plugin store" should be out of scope for this feature - support is provided only for official integrations, others are supported only provisionally.

This system does have a slight impact to XIVLauncher's security, but I would consider these changes to be a non-issue to a benefit. Password managers feel more secure than the local Windows Credential Store. Developers _may_ choose to implement bad ideas such as plugins with hardcoded OTP secrets, but this is no worse than existing integrations with the currently-existing webserver.

This system would be preferred over the webserver as it's overall simpler and can provide a better experience by removing the need for launch scripts and similar. It would also arguably be more secure, is simpler, and doesn't require firewall management.

goaaats:
It would be nice if this was a separate thing to XL and implemented as a C interface. Maybe some other tools also want to take advantage of this if we write a good spec.

PR #1651, opened 2024-12-03
OTP APK and server no more needed, fix steam not opening automatically and disabled auto update for tests

victormatuk:
Made some improvements to Steam's functionality:

- Fixed the issue with Steam not opening automatically.
- Added a configuration option to save your OTP code directly in the settings, making auto-login easier—no need for extra servers or APKs.
-  Disabled the auto-update feature for more control.

Enjoy the smoother experience! 😊


nathanctech:
I think you accidentally committed your actual XIVLauncher folder...there's the exe and all subfolders within it.

KazWolfe:
> Disabled the auto-update feature for more control.

This cannot be merged in. The auto-update system is a security feature designed to prevent the launcher from sending invalid data to SE servers, e.g. in the event of a game patch.

The warning in the title bar explains this:

```cs
Title += " - UNSUPPORTED VERSION - NO UPDATES - COULD DO BAD THINGS";
```

> Added a configuration option to save your OTP code directly in the settings [...]

This has come up before and has been shot down as it defeats the point of OTP. See #437.

If this is something you want to add, please consider working on the Authentication API as discussed in #1433.

> Fixed the issue with Steam not opening automatically.

Your fix for this seems to be increasing the retry count from 15 to 60. While this can work (I suppose), it might be better to actually look into why retries don't work. There's no real guarantee that 60 retries will be any better if there's a systemic issue that can be raised to the user instead.

---

And yes, as noted, please don't include binaries and other assets in your PR, these should be ignored properly as they're handled by the build system.

PR #1689, opened 2025-04-22
Update headers for latest boot version

NotNite:
Boot is now using WebView2, so the headers are different. Chrome 135, baby! Did not check patch-dl and friends as those aren't done through the HTML part. Tested with mitmproxy. Other observed things:

- UA computer ID is different, not sure when this changed and also not a big issue (as long as it's consistent between logins) but still something to note.
- Accept-Language was always `ja` for me, not sure why. That's written into config with the normal generated string, and AFAICT that code path should work, but something must've written `ja` into my launcher config at some point and I never noticed.
- I do not know what the numbers mean in `Accept` but I hope to god that's not something specific with my PC.
- Region should probably stop being hardcoded lol.

### DIFF
--- a/src/XIVLauncher.Common/Game/Launcher.cs
+++ b/src/XIVLauncher.Common/Game/Launcher.cs
@@ -473,12 +473,22 @@ public class Launcher
     {
         // This is needed to be able to access the login site correctly
         var request = new HttpRequestMessage(HttpMethod.Get, url);
-        request.Headers.AddWithoutValidation("Accept", "image/gif, image/jpeg, image/pjpeg, application/x-ms-application, application/xaml+xml, application/x-ms-xbap, */*");
-        request.Headers.AddWithoutValidation("Referer", GenerateFrontierReferer(this.settings.ClientLanguage.GetValueOrDefault(ClientLanguage.English)));
-        request.Headers.AddWithoutValidation("Accept-Encoding", "gzip, deflate");
-        request.Headers.AddWithoutValidation("Accept-Language", this.settings.AcceptLanguage);
+
+        request.Headers.AddWithoutValidation("Host", request.RequestUri.Host);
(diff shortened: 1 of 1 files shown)


t8d7:
> UA computer ID is different, not sure when this changed and also not a big issue (as long as it's consistent between logins) but still something to note.

I.. actually see this as a bit of a yellow flag.. as if SE reset everyones computer ID, then in their database, if they simply look for logins with a ID with timestamps before/after the change, they have a list of people to be suspect at. I believe the cautious approach is to reset (and keep consistent) in-line with the launcher.

NotNite:
Some context: There seems to be a fallback if WebView2 doesn't work (dunno if IE or CEF), theoretically this isn't *required* but it'd be nice to merge for consistency.

> actually see this as a bit of a yellow flag

I agree, it should be corrected (probably), but to SE it just looks like two PCs. This also probably wasn't introduced with this patch, just something I observed today. Something to look at in the future, but it's not high priority for the moment.

t8d7:
Fair, especially on the timing of the reset.

> Also, changing the algorithm that derives the computer ID might cause password resets for people without 2FA

That'd normally make since, except for the scenario where they officially (essentially) reset everyones computer ID, so for normal users, they see both old and new 'computers'. Considering they did it officially, it's safe to assume they halted/changed the automatic password reset system because of the change. If that holds, then us rotating IDs should be perfectly fine (if staggered, if all of XIVLauncher users suddenly change ID's, then that can also be tracked in a database (would be an annoying sql query to write though)

goaaats:
> except for the scenario where they officially (essentially) reset everyones computer ID, so for normal users, they see both old and new 'computers'. 

It didn't as far as I can tell. Nite's post is a bit ambiguous on this but I confirmed with them, they were referring to the ID XL uses being different from the one the official launcher uses which is "intentional". But yes, I agree in principle.

PR #1707, opened 2025-07-03
Add support for Encrypted DNS

YsEmei:
If you block unencrypted DNS port 53, you can't download any game patches through XIVLauncher.
Fixed by disabling Aria's built-in asynchronous DNS resolver and instead use the system's default DNS resolver.

Downloads doesn't start with async-dns enabled: 
image


YsEmei:
For those having issues with XIVLauncher forcing unencrypted DNS on the aria2c instance, add this to your C:\Windows\System32\drivers\etc\hosts 
`23.215.2.30 patch-dl.ffxiv.com`

PR #1737, opened 2025-09-06
Add a GetRawTickCount implementation for Linux

marzent:


Kobi-Blade:
Is there a specific reason this level of precision is required? Otherwise, would it be acceptable to simplify the DeriveKey method to something like the following?

```
private uint DeriveKey()
{
    var ticks = (uint)(Stopwatch.GetTimestamp() & 0xFFFF_FFFF);
    var key = ticks & 0xFFFF_0000u;

    var keyPair = new KeyValuePair<string, string>("T", ticks.ToString());
    if (arguments.Count > 0 && arguments[0].Key == "T")
        arguments[0] = keyPair;
    else
        arguments.Insert(0, keyPair);

    return key;
}
```

reiichi001:
Because it should be accurate and generate the same values as the official launcher. This is a discrepancy in how wine and dotnet (maybe also Windows) handle tracking time.

For further reading:
https://discord.com/channels/581875019861328007/850387047927250944/1413837549712965672
```
CLOCK_MONOTONIC
Clock that cannot be set and represents monotonic time since some unspecified starting point. This clock is not affected by discontinuous jumps in the system time (e.g., if the system administrator manually changes the clock), but is affected by the incremental adjustments performed by adjtime(3) and NTP.

CLOCK_MONOTONIC_COARSE (since Linux 2.6.32; Linux-specific)
A faster but less precise version of CLOCK_MONOTONIC. Use when you need very fast, but not fine-grained timestamps.

CLOCK_MONOTONIC_RAW (since Linux 2.6.28; Linux-specific)
Similar to CLOCK_MONOTONIC, but provides access to a raw hardware-based time that is not subject to NTP adjustments or the incremental adjustments performed by adjtime(3).
```

Kobi-Blade:
> Because it should be accurate and generate the same values as the official launcher.
> 
> For further reading: https://discord.com/channels/581875019861328007/850387047927250944/1413837549712965672
> 
> ```
> CLOCK_MONOTONIC
> Clock that cannot be set and represents monotonic time since some unspecified starting point. This clock is not affected by discontinuous jumps in the system time (e.g., if the system administrator manually changes the clock), but is affected by the incremental adjustments performed by adjtime(3) and NTP.
> 
> CLOCK_MONOTONIC_COARSE (since Linux 2.6.32; Linux-specific)
> A faster but less precise version of CLOCK_MONOTONIC. Use when you need very fast, but not fine-grained timestamps.
> 
> CLOCK_MONOTONIC_RAW (since Linux 2.6.28; Linux-specific)
> Similar to CLOCK_MONOTONIC, but provides access to a raw hardware-based time that is not subject to NTP adjustments or the incremental adjustments performed by adjtime(3).
> ```

Thank you for the answer. I was looking for cross-platform options, but when targeting precision timing, I couldn't find any viable alternatives yet, to make this more clean.

goaaats (approved):

PR #1772, opened 2025-11-28
Add support for fetching OTP from shell

webmakersteve:
First contribution here. Happy to make some changes.

This allows for fetching OTP from an arbitrary shell command which works great for my password manager! I am running this locally myself but would be happy to make any changes required to get this into mainline.

I know everything needs to be localized as well. I will be happy to do that if this is a change you'd accept.

### DIFF
--- a/src/XIVLauncher.Common/Shell/OtpShellReceiver.cs
+++ b/src/XIVLauncher.Common/Shell/OtpShellReceiver.cs
@@ -1,0 +1,117 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using static XIVLauncher.Common.Http.OtpListener;
+
+namespace XIVLauncher.Common.Shell
+{
+    public class OtpShellReceiver
+    {
+        private readonly string fileName;
+        private readonly string args;
+
+
+        /// <summary>
+        /// Create a shell command receiver that takes the given configured shell command
+        /// and executes it to derive the OTP
+        /// </summary>
+        /// <param name="shellCommand">The configured command. The first space position is treated as the separator between filename and args</param>
+        public OtpShellReceiver(string shellCommand)
+        {
+            // Parse shell command. Just get the first part and then the arguments afterward
+            var firstSpacePosition = shellCommand.IndexOf(" ");
+
+            if (firstSpacePosition == -1)
+            {
+                this.fileName = shellCommand;
+                this.args = string.Empty;
+            } else
+            {
+                this.fileName = shellCommand.Substring(0, firstSpacePosition);
+                this.args = shellCommand.Substring(firstSpacePosition + 1);
+            }
+        }
+
+        public async Task<(bool wasGotten, string? oneTimePassword)> TryGetOneTimePasswordAsync(CancellationToken cts = default)
+        {
+            Process proc = new Process
+            {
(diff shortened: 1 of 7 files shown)


KazWolfe:
Ref #1378 and #1433 here are related art.

I like the general concept here, and it would certainly be helpful to make MFA flows for those of us who use password managers nicer. It would be great if you could add some documentation on how to set this up, I'd think.

Otherwise, if you wanted more work in this specific area, perhaps check out the above-referenced #1433 - it might be a bit better of an overall use of time. I do have some worries that if we start adding in Too Many Systems to resolve OTP/credentials (since this is now our _fourth_ proposed mechanism), it could get a bit frustrating to maintain in the long run.

webmakersteve:
I took a look at the issue WRT adding an pluggable auth system. I thought about doing it a bit more generically when I first did this but didn't want to spend too much time if it was not something anyone was interested in - guess I should have taken a look at PRs and issues first!

The lightest weight way of doing this type of plugin system is doing it something a la protobuf plugins which just uses shell STDIN and STDOUT with a specific protocol as a way of ferrying data between processes, especially when the system is as light as this one (just literally loading account credentials). 

I'm not sure how much the app benefits from having a full pluggable system. I guess it is nice because the app wouldn't need to store any credentials (username, password, or whatever else) but I figure the only dynamic piece of auth information most people are dealing with is their OTP. 

In lieu of that, configuring the app to load all C# assemblies in something like a `plugins/` path dynamically and loading those into the classpath would be the next easiest. I can look at what i've done here and adapt it to that, and the HTTP server feature can maybe just be added to a plugin in that case.

KazWolfe:
> The lightest weight way of doing this type of plugin system is doing it something a la protobuf plugins which just uses shell STDIN and STDOUT with a specific protocol as a way of ferrying data between processes, especially when the system is as light as this one (just literally loading account credentials).

We were thinking of being fancy and having communication via a proper plugin system, but honestly even just JSON over stdin/stdout would be enough. That's more for the other thread than here; more putting out the general concept of where one can take this were they to want to go down that path. The main thought behind plugins was that we could have more complex functionality without requiring people go all out with shell scripts and the like (e.g. Yubikey support).

I do generally agree that pure OTP fetching is probably enough for the vast majority of users, but if we overengineer a _bit_, why not just go all the way? While I'm not an XL maintainer, I do think there would be interest from that side were someone to actually build a decent (and reasonably secure) solution for this.

At the very least, tentative +1 for this, just with the general concern that were we to keep adding more mechanisms, it'd become a bit annoying to deal with in the long run.

zten:
I'd avoid the overengineering bit since CLI tool integration for fetching OTP seems to be well-supported among popular tools, and it's unlikely to get any more complex than that. Or you could externalize whatever complexity you might be imagining to another CLI tool that simply outputs the OTP to stdout. (The only way this might fare poorly is if passkeys are somehow implemented for login, but Square-Enix is living in another decade entirely.)

edit: I guess you could go all the way and outsource the password management to the password manager. #1433 does seem to encapsulate that pretty well. It is a little bit more interaction with the `op` CLI than I imagined, but it does deal with the possibility that someone is using multiple accounts pretty well -- right now, the hardcoded CLI command assumes you're only fetching the code attached to one specific authenticator.  It feels less overengineered than my initial assessment.

This (or other competing CLI tool integrations) would likely be far more used than the extremely cumbersome to use macro + Square-Enix Authenticator that is already implemented in XIVLauncher, and it seems like a nice quality of life upgrade.

PR #1802, opened 2025-12-19
Fix Remote Patcher failure on Finish

sjenkins7:
This corrects the casting for the Remote Patch Installer on Finish, which triggers failures in Core.

See: https://github.com/goatcorp/XIVLauncher.Core/pull/287#issuecomment-3669901846 for context

Tested on Linux via Flatpak


wolfcomp:
This will cause the patcher to fail on windows from what has been talked about in the discord.

PR #1806, opened 2025-12-20
better detection for what platform on xlcore

wolfcomp:


wolfcomp:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation check the .NET 5+ (and .NET Core)

Contains platform symbols for Windows and OSX but not Linux so can derive it from the other two

PR #1810, opened 2025-12-22
Add Happy Eyeballs support

KazWolfe:
Ports Happy Eyeballs support from Dalamud to XL (see goatcorp/Dalamud#1187 and goatcorp/Dalamud#1189).

This PR adds the callback used over in Dalmaud, and provides a (very simplified) extended `HappyHttpClient` that functions as a drop-in replacement. It also replaces `new HttpClient()` in many places with a call to a single static shared client to avoid excessive allocations, though TBD if this is the preferred practice or not.


goaaats:
Merged by hand with some other changes, thanks!

PR #1821, opened 2026-01-06
fix: Don't root paths for free space calculations

### LINKED ISSUE #1820: Storage space check for drives mounted in NTFS folders only check the hosting drive space
### Update disclaimer

- [x] Yes, I have checked and my issue is not related to the game updating and plugins not working correctly.

### What did you do?

If one has a drive mounted in an NTFS folder like C:\Mounts\Drive, and the host drive like C:\ has less than the required size available, but the mounted drive has more than required, and if the game directory and patch directory are both on the mounted drive, the install will fail as it only checks the host drive storage size, and not the mounted drive.

### Platform

Windows

### Wine/Proton runner version

_No response_

### Relevant log output

```shell

```

KazWolfe:
`GetDiskFreeSpaceEx` allows passing in a full directory name, and will automatically resolve any reparse or volume mount points accordingly. The existing `.Root` would have always returned the base drive, thereby potentially showing invalid out of space warnings.

Linux code already passes the full path into `DriveInfo`, which will properly resolve the mount point already. For some reason, Windows' `DriveInfo` [will not properly handle mount points](https://learn.microsoft.com/en-us/dotnet/api/system.io.driveinfo.-ctor?view=net-10.0), so we still need the P/Invoke and native call.

Fixes #1820.

### DIFF
--- a/src/XIVLauncher.Common/Util/PlatformHelpers.cs
+++ b/src/XIVLauncher.Common/Util/PlatformHelpers.cs
@@ -190,7 +190,7 @@ public static class PlatformHelpers
 
         ulong dummy = 0;
 
-        if (!GetDiskFreeSpaceEx(info.Root.FullName, out ulong freeSpace, out dummy, out dummy))
+        if (!GetDiskFreeSpaceEx(info.FullName, out ulong freeSpace, out dummy, out dummy))
         {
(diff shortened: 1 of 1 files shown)


KazWolfe:
testy testy:

image

PR #1850, opened 2026-04-13
Fix language theming and release documentation

RikCost:
Made-with: Cursor


RikCost:
My mistake, I'll close it now.

PR #1867, opened 2026-05-21
Use XDG_DATA_HOME/dev.goats.xlcore for storage on Linux

rankynbass:
Includes moving ~/.{appName} to the new location, or falling back to ~/.{appName} if it can't be moved due to different drives.

The test for moving is not completely exhaustive; there's probably some weird edge case with zfs or btrfs subvolumes, or very weirdly named directories that might cause a failure, but the basic test seems to work pretty well.

This does not change the behavior on Mac OS, because I'm not sure if XIVonMac still uses this repo, and I don't want to screw things up for them.

### DIFF
--- a/src/XIVLauncher.Common/Storage.cs
+++ b/src/XIVLauncher.Common/Storage.cs
@@ -23,7 +23,7 @@ public class Storage
         else
         {
             // Keeping the old path on MacOS for now.
-            this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), $".{appName}"));
+            this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}"));
         }
 
         if (!string.IsNullOrEmpty(overridePath))
--- a/src/XIVLauncher.Common/Storage.cs
+++ b/src/XIVLauncher.Common/Storage.cs
@@ -14,14 +14,9 @@ public class Storage
         {
             this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), appName));
         }
-        else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)  || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
-        {
-            // Use XDG_DATA_HOME on Linux.
-            this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), appName));
-        }
         else
         {
-            this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}"));
+            this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), appName));
         }
 
         if (!string.IsNullOrEmpty(overridePath))
--- a/src/XIVLauncher.Common/Storage.cs
+++ b/src/XIVLauncher.Common/Storage.cs
@@ -15,14 +15,13 @@ public class Storage
         {
             this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), appName));
         }
-        else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+        else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)  || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
         {
             // Use XDG_DATA_HOME on Linux.
-            this.Root = this.getXDGStoragePath(appName);
+            this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), appName));
         }
         else
         {
-            // Keeping the old path on MacOS for now.
             this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), $".{appName}"));
         }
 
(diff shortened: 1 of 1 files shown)


src/XIVLauncher.Common/Storage.cs
```diff
@@ -9,12 +11,18 @@ public class Storage
 
     public Storage(string appName, string? overridePath = null)
     {
-        if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
marzent, line 13: We don't need really any distinction at all per platform here I think...

```new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), appName));```

should work fine and do the right thing for all three platforms.

File: src/XIVLauncher.Common/Storage.cs [hunk 2]
```diff
@@ -9,12 +11,18 @@ public class Storage
...
             this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), appName));
         }
+        else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+        {
+            // Use XDG_DATA_HOME on Linux.
+            this.Root = this.getXDGStoragePath(appName);
+        }
         else
         {
+            // Keeping the old path on MacOS for now.
```
marzent: This actually should be updated to point to `Application Support`, which `Environment.SpecialFolder.ApplicationData` already naturally does.

src/XIVLauncher.Common/Storage.cs
```diff
@@ -27,6 +35,108 @@ public Storage(string appName, string? overridePath = null)
             this.Root.Create();
     }
 
+    private DirectoryInfo getXDGStoragePath(string appName)
```
marzent: Having files being potentially moved around in `getXDGStoragePath` is a bit weird, but all of this logic can live outside of this class in xlcore as well (and is probably better there).

You can just the create the storage class (and pass in the preferred name, which can also start with `dev.goats.`, and doesn't need to be prefixed here). Then check the old folder vs the current `storage.Root` and do migration as needed.

marzent (changes requested):

src/XIVLauncher.Common/Storage.cs
rankynbass: SpecialFolder.ApplicationData is ~/.config on Linux. That's the wrong place on Linux, since we download a whole bunch of things, including the game, into the storage folder by default. Data for programs is supposed to go into ~/.local/share, and since there's so much data here, it makes more sense to use ~/.local/share if we're only using one.

I suppose we could use some sort of split storage, with GetConfigFile() and GetConfigFolder(), or change xlcore to have two storage folders, one for config and one for data.

src/XIVLauncher.Common/Storage.cs
rankynbass: You're probably right about the moving folders/storage being done on the xlcore side. I'll get rid of it here.

src/XIVLauncher.Common/Storage.cs
marzent: > SpecialFolder.ApplicationData is ~/.config on Linux.

Ah I see, you are right... Then we probably want `SpecialFolder.ApplicationData` for Windows, and `SpecialFolder.LocalApplicationData` otherwise.

> I suppose we could use some sort of split storage, with GetConfigFile() and GetConfigFolder(), or change xlcore to have two storage folders, one for config and one for data.

That would probably be cleaner in the end, but is probably not worth the effort for what is gained that way.

KazWolfe: For what it's worth, things like Dalamud's binaries, plugin installs, etc., *should* live in `Local` on Windows as well, while configs et al belong in `Roaming`. In practice this doesn't really matter since (I hope) nobody is playing FFXIV on a machine with roaming profiles, but it's a technical nitpick. It's also a bit too late to change at this point.

rankynbass: It actually looks like ApplicationData and LocalApplicationData are the same folder on MacOS: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Environment.GetFolderPathCore.Unix.cs#L101

rankynbass:
Alright, got rid of the folder moving.

src/XIVLauncher.Common/Storage.cs
```diff
@@ -9,10 +10,15 @@ public class Storage
 
     public Storage(string appName, string? overridePath = null)
     {
-        if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
         {
             this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), appName));
         }
+        else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)  || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
marzent: We don't need the extra if here

src/XIVLauncher.Common/Storage.cs
```diff
@@ -9,10 +10,15 @@ public class Storage
...
+        {
+            // Use XDG_DATA_HOME on Linux.
+            this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), appName));
+        }
         else
```
marzent, line 17: This else block can be dropped

src/XIVLauncher.Common/Storage.cs
```diff
@@ -9,10 +10,15 @@ public class Storage
...
     public Storage(string appName, string? overridePath = null)
     {
-        if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
         {
             this.Root = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), appName));
         }
+        else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)  || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+        {
+            // Use XDG_DATA_HOME on Linux.
```
marzent: The comment here can probably go too

src/XIVLauncher.Common/Storage.cs
rankynbass: Alright, dropped the else block, made the else if into else, and dropped the comment.

marzent (approved):
LGTM.

There may an argument to be had if it is better to detect platforms at runtime vs resolving that at compile time, but I think there are some issues currently with msbuild in this project in that regard already...

EDIT: Probably everything should be squashed into one commit as well.

PR #1912, opened 2026-09-12
fix: account deletion

Bronya-Rand:
For the longest time that I've used XIVLauncher, I was quite annoyed over the fact that anytime I deleted a account from it, that said account re-appeared multiple times. Deciding to look at the code myself, the literal bug apparently was that we had 2 AccountManagers that de-synced how account info got handled? I just basically told the code to use the Model's AccountManager throughout.


Bronya-Rand:
Update on this: Seems there was a new instance of AccountManager being made in the MainWindow for some reason causing a de-sync between AccountSwitcher and maybe other places too. I just passed along the Model's own AccountManager instead so that everything remained in sync with each other.

PR #1913, opened 2026-09-12
feat: add reveal password functionality

Bronya-Rand:
Simply adds the ability for users to know what they're typing in the password field. (IDK why the loc JSONs are heavily changed, I only added one field. Might be a Visual Studio thing /shrug)


MidoriKami:
It changed the indentation from 2space tabs to 4space tabs

Bronya-Rand:
Smh... Does VS really not have the spacing toggle option? Guess formatting in VSC it is.

MidoriKami:
It should be defined by the editorconfig for the FFXIVQuickLauncher project.

If it isn't, then defining it in the editor config and then applying a code cleanup to those files should fix them.

If it's not defined in an editorconfig then your IDE will use your local system/IDE settings.

Bronya-Rand:
Seems JSON wasn't included in `.editorconfig` so it defaulted to 4-space tabs. Added that so now its properly formatted.

